Issue #366, opened 2015-09-07
Brace matching take strings into account


almarklein:
Originally reported by: **Almar Klein (Bitbucket: [almarklein](http://bitbucket.org/almarklein), GitHub: [almarklein](http://github.com/almarklein))**

----------------------------------------

We now have brace matching, but it fails when you have ``foo('some ) string')``

cc @EelkeSpaak 

----------------------------------------
- Bitbucket: https://bitbucket.org/iep-project/iep/issue/366

ysalmon:
This could be solved by restricting brace-matching to portions of the code that lie out of string literals, and just skipping over string literals. Doing so can be done by reusing the logic of the code that colours the string literals.

ysalmon:
Skipping string literals is not sufficient ; comments should also be skipped.

ysalmon:
Here is a commit bb512b78ee85e69548cadaf43c91e0080280f819 adding support for parenthesis in the python parser. The next step would be to modify the HighlightMatchingBracket class to iterate over tokens instead of characters, but my understanding of how the different modules of pyzo interact is too poor for this.

ysalmon:
The current implementation is really flawed, not only because of strings and comments : it will consider ([)] as matched. Counting opening and closing parenthesis of the same kind is not sufficient, one must use a stack.

ysalmon:
I think I got it right in the updated #387 . Can you review it @almarklein ?

I modify the highlighter to store the current block's token list as a part of its userData.
This enables the brace-matcher to fetch these token lists in both directions (instead of reparsing, which would be extremely difficult when going backwards). One skips over undesired braces automatically because those are not included in the token lists.

Storing token lists in blocks may also enable some improvements in other extensions which can now dispense from reparsing parts of the document.

PR #387, opened 2016-04-08
Correcting brace matching

ysalmon:
This (in relation with #366)

- gives a correct algorithm for brace-matching,
- marks unmatched/unbalanced braces with a reddish background and
- adds parenthesis tokens in the python parser to make it possible to correctly ignore strings and comments
- stores each block's token list in its userData object.

### DIFF
--- a/pyzo/codeeditor/extensions/appearance.py
+++ b/pyzo/codeeditor/extensions/appearance.py
@@ -128,7 +128,6 @@ class _ParenIterator :
             while self.cur_tokens[k].end != cursor.positionInBlock() :
                 k += 1
             self.cur_pos = k
-            self.direction = direction
         except IndexError :
             # If the parenthesis cannot be found, it means that it is not inluded
             # in any token, ie. it is part of a string or comment
--- a/pyzo/codeeditor/extensions/appearance.py
+++ b/pyzo/codeeditor/extensions/appearance.py
@@ -115,7 +115,11 @@ class HighlightMatchingBracket(object):
     _styleElements = [  (   'Editor.Highlight matching bracket',
                             'The background color to highlight matching brackets.',
                             'back:#ccc', 
-                        ) ]
+                        ),
+                        (   'Editor.Highlight unmatched bracket',
+                            'The background color to highlight unmatched brackets.',
+                            'back:#F5A9BC',
+                        )]
 
 
     def highlightMatchingBracket(self):
@@ -138,18 +142,23 @@ class HighlightMatchingBracket(object):
         self.viewport().update()
         
     
-    def _highlightSingleChar(self, painter, cursor, width):
-            """ _highlightSingleChar(painter, cursor, width)
+    def _highlightSingleChar(self, painter, cursor, width, isMatched):
+            """ _highlightSingleChar(painter, cursor, width, isMatched)
             
             Draws a highlighting rectangle around the single character to the
             left of the specified cursor.
+            
+            The rectangle's color depends on whether the char has a corresponding parenthesis.
             
             """
             cursor_rect = self.cursorRect(cursor)
             top = cursor_rect.top()
             left = cursor_rect.left() - width
             height = cursor_rect.bottom() - top + 1
-            color = self.getStyleElementFormat('editor.highlightMatchingBracket').back
+            if isMatched :
+                color = self.getStyleElementFormat('editor.highlightMatchingBracket').back
+            else :
+                color = self.getStyleElementFormat('editor.highlightUnmatchedBracket').back
             painter.setBrush(color)
             painter.setPen(color.darker(110))
             painter.drawRect(QtCore.QRect(left, top, width, height))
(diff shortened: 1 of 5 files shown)


ysalmon:
Added iteration over tokens instead of raw text ; skips over string literals and comments correctly ; does not brace match braces that lie inside of strings or comments.

pyzo/core/editor.py
```diff
@@ -259,6 +259,7 @@ def __init__(self, parent, **kwds):
         self.setWrap(bool(pyzo.config.view.wrap))
         self.setHighlightCurrentLine(pyzo.config.view.highlightCurrentLine)
         self.setLongLineIndicatorPosition(pyzo.config.view.edgeColumn)
+        self.setHighlightMatchingBracket(pyzo.config.view.highlightMatchingBracket)
```
almarklein: Does this not crash Pyzo? I'd think that you'd also need to add a default for this in 'resources/defaultConfig.ssdf'

pyzo/codeeditor/extensions/appearance.py
```diff
@@ -108,14 +109,64 @@ def paintEvent(self, event):
...
+                self.cur_block = self.cur_block.next()
+            else :
+                self.cur_block = self.cur_block.previous()
+            if not self.cur_block.isValid() :
+                raise StopIteration
+            self.cur_tokens = self._getParenTokens()
+            if self.direction == 1 :
+                self.cur_pos = 0
+            else :
+                self.cur_pos = len(self.cur_tokens) - 1
```
almarklein: suggestion: `self.cur_pos = 0 if self.direction == 1 else len(self.cur_tokens) - 1`

pyzo/codeeditor/extensions/appearance.py
```diff
@@ -108,14 +109,64 @@ def paintEvent(self, event):
...
+            while self.cur_tokens[k].end != cursor.positionInBlock() :
+                k += 1
+            self.cur_pos = k
+        except IndexError :
+            # If the parenthesis cannot be found, it means that it is not inluded
+            # in any token, ie. it is part of a string or comment
+            raise _ParenNotFound
+    
+    def _getParenTokens(self) :
+        return list(filter(lambda x : isinstance(x, ParenthesisToken), self.cur_block.userData().tokens))
```
almarklein: Can we be certain that the `tokens` list has already been set?

pyzo/codeeditor/extensions/appearance.py
```diff
@@ -166,36 +222,36 @@ def _findMatchingBracket(self, char, cursor, doc):
...
-            if num_match == 0:
+        stacked_paren = [char] # using a Python list as a stack
+        # stack not empty because the _ParenIterator will not give back
+        # the parenthesis we're matching
+        for (token, bloc_pos) in _ParenIterator(cursor, direction) :
+            paren = token._style
+            if paren in stacking :
+                stacked_paren.append(paren)
+            elif paren in unstacking :
+                if stacked_paren == [] or self._matchingBrackets[stacked_paren[-1]] != paren :
```
almarklein: IIUC the second part of this `if` represents a mismatch. Would it be good (and feasible) to highlight (in red) both braces that make up the mismatch?

almarklein:
This is great! Made a few comments, but looks good otherwise. Will do some testing with it tomorrow.

pyzo/core/editor.py
ysalmon: It does not. I do not understand how these values are managed.

pyzo/codeeditor/extensions/appearance.py
ysalmon: Does this make a difference ?

pyzo/codeeditor/extensions/appearance.py
ysalmon: Not sure. It is set by the highlighter. Since that seems to be called when the file is loaded or modified, I think so. Maybe the block class should have a getTokenList() method which would take care of all this, instead of the highlighter.

pyzo/codeeditor/extensions/appearance.py
ysalmon: Updated the pull request with this. But I am not sure this is really interesting, so I made an option for this, defaulting to False.

pyzo/codeeditor/extensions/appearance.py
almarklein: No, it's just shorter :)  only a suggestion, feel free to leave as-is

pyzo/codeeditor/extensions/appearance.py
almarklein: We cannot really expand the Block class, because its a q Qt thing. I'd feel a bit saver if we wrapped this in a `try: ... except AttributeError: ` block.

ysalmon: Better test for tokens is None. The tokens attribute always exist, because I modified the __init__ of userData to set it to None bye default (being used to statically typed languages like OCaml, I am always very ill at ease when functions just add attributes to objects that happen to pass in their hands ; it also took me several hours to understand that the parser just adds a _style attribute to tokens when it sees fit to do so).

almarklein: Maybe init it with `[]`, so we don't have to do anything here?

ysalmon: Done.

pyzo/core/editor.py
almarklein: It makes pyzo fail during startup for me. Maybe you added it to your config somehow. It is easily resolved by adding the corresponding field to `resources/defaultConfig.ssdf`.

almarklein: Ones that is done, maybe it also makes sense to add an option to the view menu to enable/disable.

almarklein:
Sweet!

pyzo/core/editor.py
ysalmon: Yes, I added a line in my config file. Can you take care of these things relating to options/menus ? I have not read that part of the code and have no idea how it works (and do not know Qt either).
I added a method highlightMisMatchingBracket in the class for tweaking whether to highlight the red parentheses in my example ; maybe you will want to expose it to the option system too.

almarklein: Sure, I think you've done a great job here already, I can add the menu option.

almarklein:
Tested locally, all seems fine. Thank you!

Issue #392, opened 2016-04-26
Debian packaging


ghisvail:
Hi Almar, how are you doing?

I took the liberty to package `pyzo` for Debian, similar to what I did with `iep` back in the day. I added a custom `pyzo.desktop` file and placed the appdata at the right place. I am happy to provide the .desktop file for other distros to use, just let me know where you would like me to put it (under root or somewhere under pyzo/resources/ perhaps?)

Cheers,
Ghis

ghisvail:
Which icons in `pyzo` are modified / not available in [famfamfam-silk](https://packages.debian.org/sid/all/famfamfam-silk/filelist)?

ghisvail:
[Done](https://tracker.debian.org/pkg/pyzo)

ghisvail:
Please let me know regarding the .desktop file and the famfamfam-silk icons.

almarklein:
Hi Ghis, 

Thanks again, for taking this up!

Yes, `pyzo/resources` seems like a good place to put the `pyzo.desktop` file.

Icons that I modified / added:
* all icons starting with "run_" (run_cell, run_file, run_file_script, run_lines, run_mainfile, run_mainfile_script).
* all icons starting with "debug_" (debug_continue, debug_next, debug_quit, debug_return, debug_step)
* all icons starting with "overlay_" (overlay_disk, overlay_hg, overlay_link, overlay_star, overlay_svn, overlay_thumnnail)
* star2.png and star3.png
* application_erase.png and application_refresh.png
* folder_hg.png and folder_svn.png (the fact that there is no _git.png shows that its been a while since I made these :))
* page_wite_dirty.png, page_white_py.png, page_white_pyx.png, 
* plugin_refresh.png

Is this about licensing? Maybe it makes sense to prefix the custom icons to make the distinction more clear?

ghisvail:
> Yes, pyzo/resources seems like a good place to put the pyzo.desktop file.

Ok, PR to follow.

> Is this about licensing?

Yes. Plus, I should use symlinks to the icons provided by the official famfamsilk Debian package instead of the embedded ones. 

> Maybe it makes sense to prefix the custom icons to make the distinction more clear?

That would help a lot. Thanks.

almarklein:
> I should use symlinks to the icons provided by the official famfamsilk Debian package instead of the embedded ones. 

Ah, right ...

PR #397, opened 2016-05-08
Add freedesktop file for Pyzo.

ghisvail:
This is the freedesktop file I am currently using for the Debian packaging, as reported in #392.


almarklein:
Thanks!

Also cc @hmaarrfk as he is working on something related.

PR #398, opened 2016-05-09
Passed sys.argv to the QApplication function

### LINKED ISSUE #396: wmclass not set
Hi,

wmclass isn't that. That makes it impossible to allow the desktop environments to follow the application.

I don't know how to fix this. But maybe you do?

Currently it is left blank.

Mark

hmaarrfk:
Now the windows have a wmclass and can be associated to
a program in linux

Fixes #396

It seems like Qt Applications having a different look when being run under root is quite typical
https://wiki.archlinux.org/index.php/qt#Theme_not_applied_to_root_applications

I might have accidentally deleted all the trailing whitespace in the file. If you don't want those modifications, I can create a new pull request.

### DIFF
--- a/pyzo/__init__.py
+++ b/pyzo/__init__.py
@@ -41,7 +41,7 @@ Some key features
     with '##').
 
 """
-    
+
 # Set version number
 __version__ = '4.1'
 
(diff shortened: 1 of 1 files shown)


almarklein:
Thanks for the fix!

As for whitespace, Github's `?w=0` comes to the rescue :)

PR #417, opened 2016-08-02
Reduce scrolling while stepping in the debugger

lcerman:
Editor window will scroll (center the cursor) only when the cursor moves close to the window border.


almarklein:
Thanks!

lcerman:
... and thank you for the quick integration!

PR #444, opened 2016-10-01
Attempt at hiding methods

fstengel:
By piggybacking the "hide function" menu item. I use a dictionnary to
map what I want to hide to the relevant menu item. A bit
overengineered, but can be extended.

### DIFF
--- a/pyzo/tools/pyzoWorkspace.py
+++ b/pyzo/tools/pyzoWorkspace.py
@@ -270,6 +270,15 @@ class WorkspaceTree(QtWidgets.QTreeWidget):
             # Pop the 'kind' element
             kind = parts.pop(2)
             
+            # <kludge 2>
+            # the typeTranslation dictionary contains "synonyms" for types that will be hidden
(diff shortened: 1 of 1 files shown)


almarklein:
Am I correct that now it would hide methods, but not functions anymore?

fstengel:
No, it does hide both functions and methods at once or none. For instance, if in the dictionary (*typeTranslation*) one were to find the entry "codeobject": "function", then when selecting "hide functions" one would hide as well items of type "codeobject". what I do is map some object types to the types known to the filtering mechanism.

Currently my (long) code simulates something of the sort:
```python
if kind=="method":
    kind = "function"
```
The only advantge is that it is extensible, with the caveat of being a tad overengineered.

almarklein:
Ah, I see; I read it the wrong way :)

Great, its extensible, but I don't think it add too much complexity in doing so, so let's get this in!

almarklein:
Thanks!

PR #449, opened 2016-10-10
Added GUI integration support for PySide2

rreilink:
This PR adds PySide2 GUI integration

PySide2 is not yet very widespread, but is AFAIK the only LGPL solution for Qt on Python >3.4.

This PR was tested on Windows, PySide 2 installed as described here:[Compiling PySide2 from source](https://fredrikaverpil.github.io/2016/08/17/compiling-pyside2/). Furthermore, one may need to copy the Qt binary files to the Python working directory as described here:[Deploy an Application on Windows](http://wiki.qt.io/Deploy_an_Application_on_Windows)


almarklein:
Nice!

PR #462, opened 2017-01-04
correctly handling messages that *start* with carriage return

Spaak:
A line '\r blabla' should result in the previous line being wiped clean. Before this PR, only 'blabla \r' would be handled correctly (i.e. the next line would be printed on top of the 'blabla' line). With this change, also messages starting with carriage return will overwrite the previous line in the shell.


almarklein:
Thanks!

PR #466, opened 2017-01-30
run -i script.py

Diti24:
Add a menu entry to execute a script with %run -i.
Doing so is important because 
%load_ext autoreload
%autoreload 2
take effect. One can interactively change code and the changes takes effect. In the same function I added also the the change directory to script directory. The implementation is not optimal of course (I was not able to make shortcuts working). 
I think it would be very important to include this functionality in the IEP!!

See also following two discussions:
https://groups.google.com/forum/#!topic/pyzo/l3I_3AvrFdA
and
https://groups.google.com/forum/#!topic/pyzo/icfCtU_zUSw

### DIFF
--- a/pyzo/core/menu.py
+++ b/pyzo/core/menu.py
@@ -1285,6 +1285,8 @@ class RunMenu(Menu):
             icons.run_file_script, self._runFile, (True, False))
         self.addItem(translate("menu", 'Run main file as script ::: Restart and run the main file as a script.'), 
             icons.run_mainfile_script, self._runFile, (True, True))
+        self.addItem(translate("menu", 'Run as ipython ::: Run file as run -i  in ipython console.'), 
+            icons.run_file_script, self._runFileIpython)
         
         self.addSeparator()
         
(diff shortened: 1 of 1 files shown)


almarklein:
Thanks for looking into this. Do you think that we can make this work without the extra menu item? There are already so many run-options. Maybe we should use this way, when running a file while it is an IPython shell?

Diti24:
Yes, I agree. 
It could be the default way "if we have an ipython shell" when doing the "Run file as script".

almarklein:
Yes, though its not enough to check whether the shell config has the IPython flag set; It could be that IPython is not installed. So the kernel will have to communicate that to the IDE on startup.

Diti24:
Could that info perhaps get embedded in the shell object?

almarklein:
Yes. Search the code for `startup_info` and you'll see that a several information about the kernel is communicated back. It should not be that hard to add whether IPython is running.

Diti24:
Ok, I will check.
Another question I have seen that the last pyzo release is quite old (obviously depends what old means).
I was expecting that this and the other change would have been integrated quite quickly, in a official minor version change. 
What is the official relaese workflow for pyzo?
If I want to share my change with my development team, should we better work on a source code github checkout? And work directly on source? I think I have read onetime/somewhere this should be possible...right?
 
What is your general suggestion/advice in this sense?

Thanks,
Dietrich

almarklein:
There is no official release cycle, but I try to do it twice a year, so yeah, it's about time. I'm trying to find a bit of time to fix a few issues and do a release.

Yes, Pyzo works fine from source. E.g. by cloning the repo and adding that dir to your PYTHONPATH. Run with `python -m pyzo` or `python pyzolauncher.py`.

almarklein:
Actually, it seems that `shell._info['ipython']` already indicates whether ipython is loaded or not ...

almarklein:
Edit: I meant `_startup_info`.

Diti24:
I have done this, in the new branch started from 4.4.0 and  in the proposed way, in pull request:
https://github.com/pyzo/pyzo/pull/491

PR #467, opened 2017-01-31
Shell history aids

Diti24:
Improve shell history usability

### DIFF
--- a/pyzo/core/menu.py
+++ b/pyzo/core/menu.py
@@ -1390,9 +1390,17 @@ class RunMenu(Menu):
         if isStatement:
             # Get source code of statement
             code = screenCursor.selectedText().replace('\u2029', '\n').strip()
+            # add code to history
+            if code in shell._history:
+                shell._history.remove(code)
+            shell._history.insert(0,code)
             # Execute statement
             shell.executeCommand(code+'\n')
         else:
+            # add code to history
+            if code in shell._history:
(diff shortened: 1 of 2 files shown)


almarklein:
If I understand correctly, this adds snippets of code to the history also when running a selection. In principal I think this is a good idea, but I do have some questions. Does this end up as one item in the history or is the effect the same as running each line individualy? What happens if I select 100 lines? Should we perhaps show a single "item" in the history.

It also looks like you made it  such that running lines from the history itself are again added to the history. I am not sure if this works well, e.g. if a user runs multiple lines, (or one line from a multi-line item), the old items is not removed and you end up with duplicates. But then again, I never use this tool myself.

Also cc @rreilink because I think he introduced this tool.

Diti24:
Well, from my user perspective point of view I would do the whole block, because typically you want to repeat the same "action". Eventually we could put a limit in the number of lines (but probably this is not needed, I think it will be really rare that one run 100s of lines with selection.

The fron the history itself, is just a temporary workaround, because when you restart pyzo you have no history in the shell (Issue: pyzo#248). So in this way you can at least take old commands from the historyview (which contains all old commands) just the first time, and then those are availble directly with up and down arroy if you need them. For what concerns the duplicate I think this is already handled, because it removes it from the history if present. 
Once Issue: pyzo#248 this second part would not make mach sense anymore.

Issue #479, opened 2017-05-11
Feature request: Move lines up & down


hudcap:
Would love to have a shortcut for moving line(s) up and down.
The standard shortcut (e.g. in Notepad++) is ctrl+shift+{up/down}

almarklein:
Done in https://github.com/pyzo/pyzo/commit/7cad93d63d67ee94b8f5819147340a13c4950df3

PR #488, opened 2017-06-22
Update to allow moving multiple lines

hudcap:
Moves all lines in selection and keeps these lines selected so they can be moved further.
Originally created from #479


almarklein:
Much better indeed. Thanks!

Issue #282, opened 2014-03-28
Shell in separate window


almarklein:
Originally reported by: **Anonymous**

---

Hello, it would be nice to add an option to detach Shell window from main. I run multiple monitors and this would be preferable for me. Do you think such feature is feasible in the future?

---
- Bitbucket: https://bitbucket.org/iep-project/iep/issue/282

Issue #106, opened 2012-05-14
Shell widget cannot be moved on Mac


almarklein:
Originally reported by: **Almar Klein (Bitbucket: [almarklein](http://bitbucket.org/almarklein), GitHub: [almarklein](http://github.com/almarklein))**

---

_Imported from [iep.googlecode.com issue #106](https://code.google.com/p/iep/issues/detail?id=106), reported by almar.klein@gmail.com_

This has been disabled because it causes a segfault. We should fix the reason why it segfaults so we can enable moving the shell widget.

---
- Bitbucket: https://bitbucket.org/iep-project/iep/issue/106

Issue #84, opened 2012-04-17
Floating widgets (editor and shell)


almarklein:
Originally reported by: **Almar Klein (Bitbucket: [almarklein](http://bitbucket.org/almarklein), GitHub: [almarklein](http://github.com/almarklein))**

---

_Imported from [iep.googlecode.com issue #84](https://code.google.com/p/iep/issues/detail?id=84), reported by almar.klein@gmail.com_

Couple of things:
- The shell can now not be moved on the Mac because it makes IEP segfault. 
- Maybe we also want the editor a float, so that it gets a nice label?
- Maybe give each float (especially tools) a small icon in the header that drops a menu when clicked on, and allows changing the settings. This gives a consistent look and saves space (read:clutter) in the window itself. It does require overloading the default title widged ...

---
- Bitbucket: https://bitbucket.org/iep-project/iep/issue/84

PR #490, opened 2017-07-03
Allows the Shell window to float (like other tools)

snarkturne:
Allowing shell window to float is useful in dual screen mode

### DIFF
--- a/pyzo/core/main.py
+++ b/pyzo/core/main.py
@@ -167,7 +167,7 @@ class MainWindow(QtWidgets.QMainWindow):
         
         # Create floater for shell
         self._shellDock = dock = QtWidgets.QDockWidget(self)
-        dock.setFeatures(dock.DockWidgetMovable)
+        dock.setFeatures(dock.DockWidgetMovable|dock.DockWidgetFloatable)
         dock.setObjectName('shells')
         dock.setWindowTitle('Shells')
         self.addDockWidget(QtCore.Qt.RightDockWidgetArea, dock)


Diti24:
Hi, nice idea to have the shell floating.
I don't know howto but I would have a suggestion:
Wouldn't it be still more useful to have the possibility to have the single shell window floating?
I.e.: often we work with more then one shell. One is used for development, and the other for interaction or triggering processes command or whatever...
There are many use cases where having the possibility of having different independent shells (separated windows) would be useful

Diti24:
See also:
https://github.com/pyzo/pyzo/issues/282

almarklein:
@snarkturne I have a vague memory of there being problems with floating shells (on OSX?). I'll do some searching today to see what that was about, and also test this on OSX. Otherwise, this looks fine to me.

@Diti24 The idea of having multiple shells appear as separate dock widgets still makes sense, but its probably quite a bit of work. I've not yet found enough need for it to bite that bullet :P

snarkturne:
@almarklein  I'm not aware of this OSX problem. But may be it's the reason why the floating shell feature was suppressed after it had been added (last years or two years ago, I can't remember).

almarklein:
The reason why it was not allowed to be floatable seems to be that the shell widget "could get lost": https://github.com/pyzo/pyzo/commit/815b4a9305f9e3742706645b9fb996c3aa05be19. Unfortunately, I was not more specific than that.

This makes me a bit cautious for adding it back though. Perhaps its safe to (start by) allowing it to be float by a config option. People who need the feature can then turn it on, and if no problems are reported, we could consider turning it on by default.

For reference, the OS X problems that I referred to (but which seem unrelated to this specific case) are related to [this bug in Qt](https://bugreports.qt.io/browse/QTBUG-20104). See also #106 and #84.

almarklein:
I merged this PR and pushed a commit after it to only use this new behavior when the appropriate config is set. To enable, edit `appdata/pyzo/config.ssdf` or open the logger shell and write:
```
pyzo.config.settings.allowFloatingShell = 1
```

Diti24:
Hi, 
perhaps I misunderstood this feature.
I was expecting (hoping) that this configuration would have make possible to have multiple separated shell windows.
The main shell window can be detached from the IDE... but all shell (if you have more then on shell) remain in the same window.

A typical use case is:
- one shell doing monitoring and printing system status, 
- a second shell where you issue commands, 

It would be nice to have the possibility to have shell in separated windows.

I see this is still  not possible or do I simply understand how to do ?
Thanks

almarklein:
@Diti24 Qt has the notion of having float *widgets* inside a dockwidget, i.e. inside a window. Having the shell detach into its own widget is significantly more complicated.

PR #491, opened 2017-07-03
Use %run -i to run script when ipython console is used

Diti24:
I have implemented the use of %run -i, when the shell is an ipython shell


almarklein:
Awesome, thanks!

PR #501, opened 2017-08-28
make tools title translateble

flame51:


pyzo/tools/__init__.py
```diff
@@ -210,7 +210,7 @@ def loadToolInfo(self):
                     if i<0: continue
                     line = line.rstrip("\n").rstrip("\r")
                     line = line[i+1:].strip(" ")
-                    toolName = line.strip("'").strip('"')
+                    toolName = eval(line)
```
almarklein: I really don't like the use of eval unless in places where it makes sense :D

almarklein:
Thanks for this. Though I recall that there was a reason why these were not translated. I'll have to look into it better to find out what that was about. I hope to find time soon to do that, but now I'm at a conference.

flame51:
Thanks. I hope all the things on the GUI could be translateble.

pyzo/tools/__init__.py
almarklein: aah, now I realize that this is what makes the other changes possible. Not a big fan of `eval()` but I guess its use is warranted here ...

almarklein:
Thanks!

flame51:
glad to be helpful. :)

PR #507, opened 2017-08-31
Feature black and white style

flame51:
a black style.  can switch between black and white .

image


almarklein:
I appreciate the effort, but things like this should really be implemented via some sort of style system, which we do not have yet.

flame51:
yes, you are right. A style system is better.

almarklein:
With #592 we now have style files and a style editor. If you want you can still submit styles as `.theme` files.

PR #508, opened 2017-08-31
use color and parenthesis etc. to indicate class def and so on

flame51:
image


almarklein:
I like this, thanks!

almarklein:
tweaked it a little bit

PR #513, opened 2017-12-03
add replaceInAllFiles button

afourmy:
sometimes, I need to replace a pattern in all files, it's a bit long to use just replace all if there are many files

### DIFF
--- a/pyzo/core/editorTabs.py
+++ b/pyzo/core/editorTabs.py
@@ -603,7 +603,7 @@ class FindReplaceWidget(QtWidgets.QFrame):
             editor = self.parent().getCurrentEditor()
         if not editor:
             return
-        print(editor)
+
         # get current position
         originalPosition = editor.textCursor()
         
--- a/pyzo/core/editorTabs.py
+++ b/pyzo/core/editorTabs.py
@@ -193,8 +193,9 @@ class FileItem:
 class FindReplaceWidget(QtWidgets.QFrame):
     """ A widget to find and replace text. """
     
-    def __init__(self, *args):
-        QtWidgets.QFrame.__init__(self, *args)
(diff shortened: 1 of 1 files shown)


pyzo/core/editorTabs.py
```diff
@@ -193,8 +193,9 @@ def pinned(self):
 class FindReplaceWidget(QtWidgets.QFrame):
     """ A widget to find and replace text. """
     
-    def __init__(self, *args):
-        QtWidgets.QFrame.__init__(self, *args)
+    def __init__(self, editors):
+        QtWidgets.QFrame.__init__(self, editors)
+        self.editors = editors
```
almarklein: It seems that the code that instantiates this widget has not been updated yet. Note that this bit might also not be necessary, as it can use the global `pyzo.editors` instead.

afourmy: I'm not sure what you mean about the code not being updated, I replaced *args because *args is only the parent window here. I didn't know about pyzo.editors, I can use it instead. Should I revert the change back to *args for __init__ ?

almarklein: I mean that the code that creates the widget should not pass `editors`, or does it already do this? If not, I'd recommend using `pyzo.editors`.

afourmy:
Reverted back to *args and with pyzo.editors, it seems to work.

almarklein:
Awesome, thanks!

almarklein:
FYI I tweaked the interface a bit, instead of having three buttons (which take up a lot of space), there are now 2, one button to apply the replacing, and one to select the kind. That should also avoid e.g. accidentally replacing in all open files. The selection changes back to  "one" when the widges loses focus. Please let me know what you think, we can always tweak it further ...

PR #515, opened 2017-12-22
Solve confusion between open magic command and open python function

ysalmon:
Some of my students have the unpalatable habit of typing a space between the name of a function and the parenthesis, ie. ```open ("/tmp/somefile")``` instead of ```open("/tmp/somefile")```. I think this is bad style, but it is definitely syntactically valid as per [python's lexical analysis rules](https://docs.python.org/3/reference/lexical_analysis.html#whitespace-between-tokens).

The problem is that there is also a magic command with name ``open`` and pyzo gives it precedence when this word is followed by a space. This leads to very strange error messages like
```
Could not determine file name for object "("/tmp/essai")".
```
or even
```
Error in handling magic function:
  invalid syntax (<string>, line 1)
  line 347 in /opt/yann/pyzo-4.4.2/source/pyzo/pyzokernel/magic.py
  line 70 in /opt/yann/pyzo-4.4.2/source/pyzo/pyzokernel/magic.py
```
in response to ```open ("/tmp/essai", mode="r")```.

I see that the magic.py code has some quirks to verify that the command is not a identifier in the current or global scope. But it only does this for a magic command without argument, so that evaluating ```open``` correctly gives ```<function io.open>```.

You have to restrict this disambiguation to commands without arguments because if you did not, it would never be possible to use the ```open``` magic command.

However, this is problematic. Renaming the ```open``` magic command would be of some help, but ultimately one cannot prevent a situation where the new name also happens to be used as an identifier in the python program.

The only solution I can see is to have magic commands with names that are not valid python identifiers.

I propose that magic command be prefixed with a ```%``` (except those which already have a ```?```). This breaks backward compatibility somewhat, but I assume people are not maintaining scripts consisting of pyzo magic commands.

### DIFF
--- a/pyzo/pyzokernel/magic.py
+++ b/pyzo/pyzokernel/magic.py
@@ -15,6 +15,8 @@ import os
 import re
 import time
 import inspect
+import tokenize, token, keyword
+import io
 
 
 # Set Python version and get some names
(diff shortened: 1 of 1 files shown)


almarklein:
I see the problem and agree that we should do something about it.

May it be as simple as testing that the next character is a letter, number, or quote? Such a thing (a variable name followed by a space and then another variable name or literal) is invalid Python syntax, so this should take care of the ambiguity.

ysalmon:
Tried a new method : using python's tokenizing routines and ensuring that the second relevant token, if any, is not an operator (which includes opening parentheses).

pyzo/pyzokernel/magic.py
```diff
@@ -108,23 +110,43 @@ def _convert_command(self, line):
...
         
-        # Check if it is a variable
-        command = line.rstrip()
-        if ' ' not in command:
-            if command in interpreter.locals:
+        # Check that line is not some valid python input
+        try :
+            # gets a list of 5-tuples, of which [0] is the type of token and [1] is the token string
+            ltok = list(tokenize.tokenize(io.BytesIO(line.encode('utf-8')).readline))
+        except tokenize.TokenError :  # typically this means an unmatched parenthesis
```
almarklein: I appreciate the effort to use the tokenize function. Unfortunately, this will fail on 2.x, where there is no BytesIO (at least not on 2.6 and below), and `tokenize()` works differently too.

ysalmon: Ah. According to the [Python2.6 doc](https://docs.python.org/2.6/library/tokenize.html), there is still some hope with StringIO instead (see example). But as I am completely incompetent in Python2, I just added a version check and fell back to the old behaviour for Python2 (maybe this can be refined to some Python2.x).

Isn't Python 2.6 outdated anyway ?

almarklein: >  I am completely incompetent in Python2

I can only applaud that :)

Nevertheless, one of the ideas of Pyzo is that you can run any Python interpreter on your system. Also Pypy and Jython. And also old versions: it goes as low as Python 2.4 :D

almarklein:
One last request: can you factor out the token-related code into a separate function? When I have a free slot I will just merge it and make it work (or not-break :)) on Python 2.x.

ysalmon:
I think that after my latest commit, the patch is neutral w/r previous behaviour for Python 2.

Do you think a 4.4.3 version with this patch could be released in a relatively near future ?

almarklein:
Thanks. I hope to find some time to do a bit of work on Pyzo and do a release soon, though atm I am rather busy. What's "relatively near future" for you? :)

ysalmon:
It is a time before the moment when people from the Région ask me « Which version of Pyzo do you want us to deploy on all the computers ? ». Which in turn is quite indeterminate.

almarklein:
About when will they ask you this question? I cannot promise anything, but I might consider it in setting my OSS priorities :)

ysalmon:
It should be within 3 weeks now.

almarklein:
Thanks! I needed to tweak it just a bit to make it work on 3.6 and pypy. I'm trying to release this week or early next week.

almarklein:
@ysalmon a new release is available, you owe me a beer :)

ysalmon:
Thanks a lot ; we should be starting deployment next week !

If you happen to be near Tours one day…

PR #518, opened 2018-01-14
Add toggle case and printify in Edit menu

kelsa-pi:


almarklein:
@kelsa-pi Thanks for the contribution!

I think the toggle case command is a valuable addition, but I am less sure about the printify action, to be honest. It targets a very specific use-case, and IMO it is not worth the extra menu item. Someday, we'll have a plugin system that should allow specific actions such as this :)

kelsa-pi:
@almarklein 
Toggle case command as new menu item is OK  :)

kelsa-pi:
I hope it's OK.

pyzo/core/menu.py
```diff
@@ -1,4 +1,4 @@
-# -*- coding: utf-8 -*-
++# -*- coding: utf-8 -*-
```
almarklein: ?

pyzo/codeeditor/base.py
```diff
@@ -867,4 +867,32 @@ def updateMargins(self):
         accordingly.
         """
         leftmargin = self.getLeftMargin()
-        self.setViewportMargins(leftmargin , 0, 0, 0)
+        self.setViewportMargins(leftmargin , 0, 0, 0
```
almarklein: This looks like a syntax error.

almarklein:
Thanks, I added two minor comments.

almarklein:
If you address the two minor comments, I can merge this before the next release :)

almarklein:
I just went ahead and merged this. I fixed the two minor issues in a new commit. Thanks for this contribution!

PR #519, opened 2018-01-21
Add builtins, instance and constant in style

kelsa-pi:
Before

before

After

after

### DIFF
--- a/pyzo/codeeditor/base.py
+++ b/pyzo/codeeditor/base.py
@@ -189,7 +189,7 @@ class CodeEditorBase(QtWidgets.QPlainTextEdit):
         blue    = "#268bd2"
         cyan    = "#2aa198"
         green   = "#859900"  # noqa
-        
+
         if True: # Light vs dark
             #back1, back2, back3 = base3, base2, base1 # real solarised
             back1, back2, back3 = "#fff", base2, base1 # crispier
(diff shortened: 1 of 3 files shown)


almarklein:
Hi @kelsa-pi, I really appreciate the effort, but I have to be honest and "slow you down" a bit :)

Constants are variables like any other and are (by convention) already distinguishable by being uppercase.

Same for builtins. E.g. I would not like `dict` be colored different from `OrderedDict` just because it's a builtin.

My initial take on `self` is the same. It's only by convention that it's called such. However, it's a very strong convention, and I can see how discerning it a bit might help. So I would accept a patch for that. But please make it more subtle, e.g. make it italic.

kelsa-pi:
Hi,
I agree with you regarding the constants, case is enough.

" I would not like dict be colored different from OrderedDict just because it's a builtin."
I am not saying it should be, but I would like to at least have the possibility to change it, if need be. Same goes for self.

Changes presented above are exaggerated, just to make a point. My proposal: add the possibility to change highlight for builtins and self, but leave the font and formatting as it was before. Nothing changes for the majority of users, but some may find it useful.

Best regards

almarklein:
Thanks for the explanation. Agreed. If you can remove the constants-part, then we can merge this.

I was planning on releasing today, but I guess I can postpone that 1 or 2 days, so if you can find the time to make the change before Friday, I can get these changes into the next release.

kelsa-pi:
I hope is OK now.
Best regards

PR #520, opened 2018-01-26
Do not show file not found error dialog at startup

FedeMiorelli:
When using a portable installation (appdatadir = settings) it is often the case that file paths are not found on the next run, due to the program being executed on a different machine.
This results in an inconvenient series of error dialogs at startup.

This patch will first check existence of files and open only the available ones. If no file is found at all, a new file is created.


almarklein:
Nice catch! Ready to go?

almarklein:
Merged. Thanks!

FedeMiorelli:
Great, thank you!

PR #530, opened 2018-03-02
replace async per basync to become Python-3.7 compatible

stonebig:


stonebig:
hum, it's no stupidly sufficient, but now it goes up to the splash creeen.

stonebig:
I see ansible project decided to pick async_val as the replacement to async

stonebig:
maybe it's sufficient, in fine:

pyzo-3 7

almarklein:
Nice, thanks! Is this ready?

stonebig:
Yes, given the cosmetic change, and the proven result.
Yet, I'm not an usual end-user, so I only pushed a few buttons and launch a few shells.

PR #544, opened 2018-06-29
Automatic insertion of quotes, braces and brackets

kelsa-pi:
Basic automatic insertion of quotes, braces and brackets


almarklein:
Thanks! This is really nice!

Can we also add it for parenthesis `()`? Actually, I think that it would be intuitive to do it for all types of brackets (`[{()}]`), or not. So IMO there should be two new menu options: "Auto close brackets" and "Auto close quotes".

A point that is worth discussing is whether we want some sort of "soft" behavior, so that this feature does not sit in the way for someone who still closes the quotes manually. The advantage of this is that we can turn this feature on by default, which would be great, because otherwise I suspect most users wont ever turn it on.

In the Jupyter notebook, for instance, this is works as follows (http://jupyter.org/try): If you press e.g. `]`, and the next char is a square bracket already, then the cursor simply skips over it, without inserting a new character. Same for quotes. I am not 100% if this is the best approach, but I've never found it annoying (that said, I rarely use it).

I'll look into how other tools do this, and maybe ask around a bit with more regular Notebook users.

almarklein:
I just tried with Sublime and VS Code. They all do it in the same way as the Notebook. Let's follow that approach!

kelsa-pi:
Hi,
Please, ignore changes in __init__.py - this is for later :)
Regards

almarklein:
np, just ping me when this is ready for review (or when you need help).

kelsa-pi:
Now is ready for review
Best

pyzo/resources/defaultConfig.ssdf
```diff
@@ -56,7 +56,8 @@ settings= dict:
     removeTrailingWhitespaceWhenSaving = 0
     changeDirOnFileExec = 0
     allowFloatingShell = 0
-
+    autoClose_Quotes = 0
```
almarklein: Please set these to 1 :)

pyzo/core/menu.py
```diff
@@ -1788,6 +1797,17 @@ def _setMode(self, autocompmode):
     def _setCompleteKeywords(self, value):
         pyzo.config.settings.autoComplete_keywords = bool(value)
 
+    def _setCompleteKeywords(self, value):
```
almarklein: You accidentally copied this method

pyzo/codeeditor/extensions/behaviour.py
```diff
@@ -441,3 +442,90 @@ def _paste(self, keepSelection):
...
+                # triple-quoted
+                cursor2 = self.textCursor()
+                textBeforeCursor = ustr(cursor2.block().text())[:cursor2.positionInBlock()]
+                if textBeforeCursor.endswith(chr(event.key())*3):
+                    cursor2.insertText(insert_txt)
+                    self._moveCursorLeft(2)
+        else:
+            super().keyPressEvent(event)
+
+    def _getNextCharacter(self, cursor):
```
almarklein: It's probably good to use double underscore for these method names to avoid potential name clashes with other editor plugins.

pyzo/codeeditor/extensions/behaviour.py
```diff
@@ -441,3 +442,90 @@ def _paste(self, keepSelection):
...
+            cursor.insertText(insert_txt)
+            self._moveCursorLeft(1)
+        
+        # closing brackets
+        elif event.key() in closeBrackets and pyzo.config.settings.autoClose_Brackets:
+            idx = closeBrackets.index(event.key())
+            next_character = self._getNextCharacter(cursor)
+            if next_character:
+                # skip
+                if ord(next_character) in closeBrackets:
```
almarklein: Should this be `if ord(next_character) in event.key():` or similar?

As it is now, If I am in from of *any* closing bracket, typing any closing bracket will skip over it; it should only work when the match is exact.

almarklein:
Nice! It already seems to work very well. Just a few comments.

almarklein:
Awesome! The tripple-quoting appears to only work when tried at the last line of a file, but it's easily fixed. I will just merge this now and fix it in master :)

Thanks for the work, this is a really great addition!

almarklein:
Fixed tripple quoting in 2d4d70ec5395f00ec0654b64e33afa8a236d379a

Also added some logic related to selected text e973b37c15281c1fb59bde64e3575fc68f99432b

(I copied the behavior from MS Code)

almarklein:
For the record, I just made some changes to make this functionality sit less in the way, again, looking at MS Code for when *not* to autoclose. Summary, not when in a comment or string. Quotes not when next to an identifier. Braces only when there's whitespace to the right.

almarklein:
Somewhat more tweaking. I got rid of the auto-tripple-close quoting, because it can also be very anoying, doing more harm then good. I realize that I was the one asking for that feature, sorry for the time you spend on that ... That last feature is present in the notebook but not in VS code, and for good reason, I think.

almarklein:
(This has been a huge feature, like, a lot of work to get right, but also hugely useful! Thanks again!)

kelsa-pi:
Thank you for taking the time to work to get it right.

kelsa-pi:
Thank you for taking the time to improve this feature.

PR #549, opened 2018-08-03
Sort in the Source structure plugins

kelsa-pi:


almarklein:
This PR would need a bit of work: when the source changes, the original order is applied again. I think you'd want orders "original", "a-z", "z-a". You probably want case independence?

That said, after considering this feature, I am declining it. I believe that the source structure should have a strong correspondence to the source code, and changing the order of the elements breaks down this correspondence. Please don't be discouraged or take this personally! I like seeing people doing things like this with Pyzo :) However, I think its important that Pyzo keeps following its key principles, one of which is simplicity.

You can always create your own version of the source structure! E.g. by placing them in `os.path.join(pyzo.appDataDir, 'tools')`.

kelsa-pi:
@almarklein 
Thank you for taking the time to consider this feature :)
>> I think you'd want orders "original", "a-z", "z-a".
"orginal"  is OK in most cases, 
"a-z", "z-a" is IMHO must for navigation/copy-paste in complex cases (eg. menu.py)
If the source changes, and the original order is applied again, one has  "a-z", "z-a" option for navigation if needed.
Best

PR #547, opened 2018-07-18
Automatic insertion of colons after class, def, if...

kelsa-pi:

Issue #235, opened 2013-09-03
New syntax parser system


almarklein:
Originally reported by: **Almar Klein (Bitbucket: [almarklein](http://bitbucket.org/almarklein), GitHub: [almarklein](http://github.com/almarklein))**

---

We currently do the syntax parsing ourselves. This works, but we only support Python and a bit of C. If we use Pygments, we can support many more languages. 

Further, the syntax parsing should be combined with sourse stucture parsing and the source structure tool should use this information to display the source structure in the same colors.

Further, this information can be used to provide a pylint like bar, and we can parse for PEP8 complience.

So in all, this overhaul will result in a lot of nice improvements.

---
- Bitbucket: https://bitbucket.org/iep-project/iep/issue/235

PR #550, opened 2018-08-06
Automatic insertion of colons after class, def, if...

kelsa-pi:
New PR, replace for (#547)
I am sorry for the inconvenience.

### DIFF
--- a/pyzo/codeeditor/extensions/behaviour.py
+++ b/pyzo/codeeditor/extensions/behaviour.py
@@ -602,3 +602,27 @@ class AutoCloseQuotesAndBrackets(object):
         cursor2 = self.textCursor()
         cursor2.movePosition(cursor2.Right, cursor2.MoveAnchor, n)
         self.setTextCursor(cursor2)
+
+
+class AutoInsertColons(object):
+    """
+    Automatic insertion of colons after class, def, if...
+    """
+
+    def keyPressEvent(self, event):
+
+        if event.key() in (Qt.Key_Enter, Qt.Key_Return):
+            if pyzo.config.settings.autoInsert_Colons:
+
+                cursor = self.textCursor()
+                textBeforeCursor = ustr(cursor.block().text())[:cursor.positionInBlock()]
+                textBeforeCursor = textBeforeCursor.strip()
+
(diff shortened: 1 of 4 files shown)


pyzo/codeeditor/extensions/behaviour.py
```diff
@@ -602,3 +602,27 @@ def _moveCursorRight(self, n):
...
+
+
+class AutoInsertColons(object):
+    """
+    Automatic insertion of colons after class, def, if...
+    """
+
+    def keyPressEvent(self, event):
+
+        if event.key() in (Qt.Key_Enter, Qt.Key_Return):
```
almarklein: This is Python specific, so we should add: `if 'python' in self.parser().name().lower():`

almarklein:
> I am sorry for the inconvenience.
No problem!!


This looks good, and I expect it would work just fine in 95% of cases. But in order to make it work correctly in the remaining 5% this needs (a lot?) more work .... imagine for example:

```py
if (something == 42 and <enter because I want the rest of the test on a new line>

if (something == 42 and
    somethingelse == 7 and
    thatotherthing != 3) <enter, NOW we must insert a colon!>

def foo():  # we should rename this <enter, we need to consider comments and whitespace>

```

The best way to do this is to run through the tokens, see https://github.com/pyzo/pyzo/blob/master/pyzo/codeeditor/extensions/behaviour.py#L454 for an example. The second case is especially tricky; we may need to look at previous lines, possibly more than one. But how would we know that the current line is part of an if/def/etc.?

Doing this correctly requires higher level information then we have now. It might be that if source code parsing (as done by the source structure tool) is combined with the code-editor (#235), this information *is* available. Until then, I think this feature will be very hard to get right, unfortunately.

kelsa-pi:
OK, I'm working on it

ysalmon:
If this gets merged, please do add an option to turn it off like for parentheses autocompletion, as this is something that I surely would not want with students (this would hinder their learning of the syntax).

almarklein:
Closing this. I suspect that an option like this would easily become annoying rather than helpful :)

@kelsa-pi You're free to try and prove me wrong :D  but it's probably good to know I'd be reluctant before putting a lot of enery into this :)

PR #556, opened 2018-08-12
Sniffshebang

ysalmon:


pyzo/codeeditor/manager.py
```diff
@@ -212,7 +215,50 @@ def suggestParserfromFilenameExtension(cls, ext):
...
+    def suggestParserfromText(cls, text) :
+        """ suggestParserfromText(text)
+        
+        Given a text, returns the name of the suggested
+        parser corresponding to the language of the file.
+        
+        See also registerShebangKeyword()
+        """
+        shebangline = None
+        for line in text.splitlines()[:5] : # is there a simple way to not read all of text ?
```
almarklein: Maybe just truncate it, e.g. `text[:999].splitlines()[:5]`. That should resolve your concern for huge files and still be a large enough value to cover the first 5 lines in 99.9% of cases.

almarklein:
Added a small comment to a question you posed in the code. Other than that, is this ready?

ysalmon:
Yes it is ; I have been running pyzo with it since I opened the PR.

almarklein:
Thanks for this nice addition!

PR #563, opened 2018-10-08
Update guiintegration.py

Diti24:
QtWebEngineWidgets needs to be imported before QCoreApplication is created otherwise it generates the following error:
ImportError: QtWebEngineWidgets must be imported before a QCoreApplication instance is created


almarklein:
There'd need to be a `try-except` around this, because `QtWebEngine` is only available on later versions of Qt5, and may be omitted on some builds.

But even then ... `QtWebEngineWidgets` is incredibly heavy, and I am very reluctant to import it for every user ... sorry :/

PR #564, opened 2018-10-10
Update interpreter.py

Diti24:
Make the
#After_GUI tag in the start-up script be a "splitter" so it is possible to distinguish between code to run Before and After the GUI creation

### DIFF
--- a/pyzo/pyzokernel/interpreter.py
+++ b/pyzo/pyzokernel/interpreter.py
@@ -450,7 +450,6 @@ class PyzoInterpreter:
                 self.context._stat_interpreter.send('Busy')
                 msg = {'source': scriptBefore, 'fname': '<startup>', 'lineno': 0}
                 self.runlargecode(msg, True)
-                
         elif script and os.path.isfile(script):
             # Run script
             self.context._stat_interpreter.send('Busy')
--- a/pyzo/pyzokernel/interpreter.py
+++ b/pyzo/pyzokernel/interpreter.py
@@ -434,7 +434,7 @@ class PyzoInterpreter:
         # Should we use the default startupScript?
         if script == '$PYTHONSTARTUP':
             script = os.environ.get('PYTHONSTARTUP','')
-            
+        
         if '\n' in script:
             # Run code later or now
             linesBefore = []
--- a/pyzo/pyzokernel/interpreter.py
+++ b/pyzo/pyzokernel/interpreter.py
@@ -437,13 +437,20 @@ class PyzoInterpreter:
         
         if '\n' in script:
             # Run code later or now
-            firstline = script.split('\n')[0].replace(' ', '')
-            if firstline.startswith('#AFTER_GUI'):
-                self._codeToRunOnStartup = script
+            scriptSplit = script.split('\n')
+            foundAfter = False
+            for idx, line in enumerate(scriptSplit):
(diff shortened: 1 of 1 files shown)


pyzo/pyzokernel/interpreter.py
```diff
@@ -437,13 +437,20 @@ def _run_startup_code(self, startup_info):
...
+                self._codeToRunOnStartup = '\n'.join(scriptSplit[idx::])
+                runBefore = '\n'.join(scriptSplit[0:idx])
             else:
-                self.context._stat_interpreter.send('Busy')
-                msg = {'source': script, 'fname': '<startup>', 'lineno': 0}
-                self.runlargecode(msg, True)
+                runBefore = script
+            self.context._stat_interpreter.send('Busy')
+            msg = {'source': runBefore, 'fname': '<startup>', 'lineno': 0}
+            self.runlargecode(msg, True)
```
almarklein: Good start! I'd like these last three lines to run only when there is actually any code to run *before*. Something like this, for instance:
```py
linesBefore = []
linesAfter = script.splitlines()
while linesAfter:
    if linesAfter[0].startswith('#AFTER_GUI'):
        break
    linesBefore.append(linesAfter.pop(0))
scriptBefore ='\n'.join(linesBefore)
scriptAfter='\n'.join(linesAfter)

if scriptBefore.strip():  # don't trigger when only empty lines
    ...
 ```

**Comment by almarklein** [comment]
Thanks!

Could you please also change the text after `#AFTER_GUI` in shellconfig.py (I think) that explains the behavior?

**Comment by Diti24** [comment]
Done in 
https://github.com/pyzo/pyzo/pull/565

File: pyzo/pyzokernel/interpreter.py [hunk 2]
```diff
@@ -437,13 +437,20 @@ def _run_startup_code(self, startup_info):
         
         if '\n' in script:
             # Run code later or now
-            firstline = script.split('\n')[0].replace(' ', '')
-            if firstline.startswith('#AFTER_GUI'):
-                self._codeToRunOnStartup = script
+            scriptSplit = script.split('\n')
+            foundAfter = False
+            for idx, line in enumerate(scriptSplit):
+                if line.replace(' ', '').startswith('#AFTER_GUI'):
```
almarklein: Small suggested tweak to make the code faster; the script is potentially very large.
```suggestion
                if line.startswith(('#AFTER_GUI', '# AFTER_GUI')):
```

almarklein: Oh wait, this is only the startup script, which we can assume to be small. Nevermind. I need a coffee.

pyzo/pyzokernel/interpreter.py
almarklein: Sorry for almost completely rewriting your contribution. 

It would be sufficient to only put the three lines above in a ``if scriptBefore.strip(): ``

almarklein:
Great, thanks for this nice addition!

almarklein:
Yes, it will be included. I understand that it has been a while. It would be good to have a release in a month or so. But I am juggling my "OSS time" between different projects, so I can't promise anything.

Note that you can always patch your own copy of Pyzo (also the binary version, see its source folder). Or you can clone this repository and run it from source.

PR #570, opened 2018-10-27
Update menu.py

kelsa-pi:
Add tools description in tooltip


almarklein:
Unfortunately, this wont make the translations work. The translation mechanism examines the source code, looking for `translate()` calls. But the text must be the the string string definition for it to work ...

Issue #587, opened 2019-02-26
Syntax Schemes; themes editor and management


erik-helmers:
Hello there !
As many people ask for a way to edit color scheme  (#80 , #306) , I tried to help a bit,
and I came up with this solution : 
      
- Themes loading in SSDF format (loads *.theme files). I'm not sure this is the best format maybe (INI or YAML maybe best) but I kept it to stay coherent with the config files.
- The active theme is loaded in the codeeditor/base.py
- I've done a little theme editor in core/menu.py

It allows to have theme files in the user directory, as well as the pyzo dir.

![proof of concept](https://user-images.githubusercontent.com/24685566/53383930-819eba00-3979-11e9-90d5-566cf5a5ffc1.png)

A bit like in a IDLE fashion, when you click on some text, you jump to the style element you want to edit. You can change colors with QColorDialogs, and see as you go the changes you're doing.
The themes located in pyzoDir/resources are not editable, you have to create a new, custom theme to modify it (the comboBox at the top is editable only when the theme is not builtin).

Anyway there's probably a lot to improve on this, i.e. when you exit the dialog, and want to apply a new theme, there's simlpe a call to `pyzo.main.restart()` (maybe there's a proper way but I don't have searched much for the moment) and the StyleElement names are not very pretty. 

PS : thank you for your hard work, this is the best Python IDE :)

almarklein:
Done in #592 / #588

PR #588, opened 2019-02-26
Adding a theme editor

erik-helmers:
Hello again !
Sorry, i'm a bit of a git noob, so this is in continuation of #587 
I hope it's clear and documented enough.

There's maybe things to improve in the look, but that's the best I could do.


almarklein:
This looks awesome, thanks a lot! I will have a proper look at this soon!

pyzo/__init__.py
```diff
@@ -170,6 +170,27 @@ def resetConfig(preserveState=True):
     _saveConfigFile = False
     print("Deleted user config file. Restart Pyzo to revert to the default config.")
 
+def loadThemes():
+    """
+    Load default and user themes (if exist)
+    """
+    
+    def loadThemesFromDir(dname, isBuiltin=False):
+        for fname in filter(lambda fname: fname.endswith(".theme"), os.listdir(dname)):
```
almarklein: I added a check that the dir exists here

pyzo/codeeditor/base.py
```diff
@@ -170,89 +169,11 @@ def __init__(self,*args, **kwds):
...
-#         S["Editor.Highlight current line"] = "back: light grey"
-#         S["Editor.Indentation guides"] = "linestyle:solid, fore: light grey"
-#         S["Editor.Line numbers"] = "back: light grey, fore: black"
-
+        
+    ## Style import settings
+        import pyzo
         # Apply style
-        self.setStyle(S)
+        self.setStyle(pyzo.themes[pyzo.config.settings.theme]["data"])
```
almarklein: I kept the default style here. Mostly because this codeeditor submodule was designed to be used separate from Pyzo. So the line above can also not be used. I moved the logic for setting the theme from Pyzo's config to `editor.py`.

pyzo/core/menu.py
```diff
@@ -1888,7 +1887,460 @@ def _selectLanguage(self, languageName):
         m.setIcon(m.Information)
         m.exec_()
 
+## Classes to enable editing the color scheme
```
almarklein: I moved this lot to a new file.

almarklein:
I took this and tweaked it a bit in #592. Thanks a lot!

pyzo/codeeditor/base.py
erik-helmers: You're right ! But there's a problem when setting the theme in `editor.py`,  (at startup the theme is not applied correctly on already opened files). The easiest fix is to put the code you wrote :
```suggestion
       # Set style/theme
        try:
            self.setStyle(pyzo.themes[pyzo.config.settings.theme.lower()]['data'])
        except Exception as err:
            print("Could not load theme: " + str(err))
```
in `baseTextCtrl.py`. It also apply the theme to the shell which is nice.

almarklein: Oh, good idea!

almarklein: done!

Issue #575, opened 2018-11-05
"edit shell configuration" doesn't seem to work on my PC


stonebig:
well, it doesn't work even going back to pyzo-4.5.1 or to current master ... What could be a likely cause of this regression on my new PC ?

almarklein:
Try removing (or better, renaming) the config file (see help > about for the application directory where the config file is).

almarklein:
Odd ... did you recently install a new Python interpreter (of any kind)?  Pyzo tries to detect interpreters on your system when the shell config is loaded. It also tries to retrieve their versions, so something may be going wrong there ... Have you tried opening the logger shell to see if any errors are produced?

almarklein:
(or running Pyzo from a terminal to capture any stdout)

stonebig:
ok, will digg along these 2 ideas. thanks, Almar.

hudcap:
I just started having this problem after installing a standard Python 3.7.2 environment alongside my Miniconda environment on Windows 10.
Pyzo version 4.7.1
Logger shows the following:

```
Uncaught Python exception: _getfullpathname: path should be string, bytes or os.PathLike, not NoneType
  File "C:\Program Files (x86)\pyzo\source\pyzo\core\menu.py", line 296, in <lambda>
    a.triggered.connect(lambda b=None: callback())
  File "C:\Program Files (x86)\pyzo\source\pyzo\core\menu.py", line 1088, in _editConfig2
    d = ShellInfoDialog()
  File "C:\Program Files (x86)\pyzo\source\pyzo\core\shellInfoDialog.py", line 607, in __init__
    self.interpreters = list(reversed(get_interpreters('2.4')))
  File "C:\Program Files (x86)\pyzo\source\pyzo\util\interpreters\__init__.py", line 27, in get_interpreters
    pythons = _get_interpreters_win()
  File "C:\Program Files (x86)\pyzo\source\pyzo\util\interpreters\__init__.py", line 89, in _get_interpreters_win
    found = [os.path.normcase(os.path.abspath(v)).strip('\\') for v in found]
  File "C:\Program Files (x86)\pyzo\source\pyzo\util\interpreters\__init__.py", line 89, in <listcomp>
    found = [os.path.normcase(os.path.abspath(v)).strip('\\') for v in found]
  File "C:\Python37\lib\ntpath.py", line 526, in abspath
```

[...]

PR #596, opened 2019-03-24
workaround badly or unregistered python

stonebig:
solves https://github.com/pyzo/pyzo/issues/575


almarklein:
Great! Thanks for digging into this!

PR #607, opened 2019-05-16
Update and complete German translation

be-we:
Update and complete German translation


almarklein:
Awesome, thanks!

be-we:
Pyzo is worth the effort. :wink:

PR #613, opened 2019-07-02
Disambiguate Python 2 and 3

ysalmon:
Currently, there is a generic Python parser that has the reunion of the keywords of Python 2 and Python 3. 
In fact this parser is used more frequently than either the specific Python 2 and Python 3 parser due to the detection process.

This is problematic because there does not actually exist a generic Python language. Especially, this generic-python parser will typeset ``print`` as a keyword (like Python 2) although it really is just a function in Python 3. This is misleading for students, who sometimes struggle to understand the difference between a keyword and an identifier, as well as between the semantics of ``print`` and ``return``.

My proposal is to explicitly label such cases as _ambiguous_ rather than _generic_, to try to limit such cases, and, at least in the case of a newly created file, default to Python 3.

To achieve this, I first prioritise the shebang analysis I introduced last year over the filename analysis (in line with the usual unix practices).

I then transform the PythonParser base class into a class that has no keyword instead of the reunion of Python2 and Python3 (we might use the intersection of these keywords sets), add an AmbiguousPythonParser class to serve the old purpose of PythonParser.

I also introduce a mechanism by which a Parser has a [``disambiguate``](https://github.com/pyzo/pyzo/pull/613/files#diff-6f3fb159ab8a06faab8afd3fb5f9ddf2R81) class method taking the text of the code and where we could use rules of thumb to choose between Python2 and Python3. This is the WIP part, for now [this method](https://github.com/pyzo/pyzo/pull/613/commits/dc30bcab544e1ee3cb6124fe59bd36480b88702c#diff-77762366107855b600578ac03a3cf662R417) is just the identity.

This requires a bit of back-and-forth between parser classes and parser names in [``suggestParser``](https://github.com/pyzo/pyzo/pull/613/files#diff-77a55db0c20e5d404787f49d717f98f1R266).

What is your opinion on this ?

### DIFF
--- a/pyzo/codeeditor/manager.py
+++ b/pyzo/codeeditor/manager.py
@@ -115,11 +115,7 @@ class Manager:
         # Put in list with the parser names as keys
         parserInstances = {}
         for parserClass in foundParsers:
-            name = parserClass.__name__
-            if name.endswith('Parser') and len(name)>6:
-                
-                # Get parser identifier name
-                name = name[:-6].lower()
+                name = cls.getParserName(parserClass)
                 
                 # Try instantiating the parser
                 try:
@@ -139,6 +135,15 @@ class Manager:
         # Store
         cls._parserInstances = parserInstances
     
+    
+    @staticmethod
+    def getParserName(parserClass) :
+        name = parserClass.__name__
+        if name.endswith('Parser') and len(name)>6:
+            
+            # Get parser identifier name
+            name = name[:-6].lower()
+        return name
     
     @classmethod
     def getParserNames(cls):
--- a/pyzo/codeeditor/manager.py
+++ b/pyzo/codeeditor/manager.py
@@ -260,9 +260,9 @@ class Manager:
         
         See also registerFilenameExtension() and registerShebangKeyword()
         """
-        parser = cls.suggestParserfromFilenameExtension(ext)
+        parser = cls.suggestParserfromText(text)
         if parser == "" :
-            parser = cls.suggestParserfromText(text)
+            parser = cls.suggestParserfromFilenameExtension(ext)
         return parser
     
     @classmethod
--- a/pyzo/codeeditor/manager.py
+++ b/pyzo/codeeditor/manager.py
@@ -115,7 +115,7 @@ class Manager:
         # Put in list with the parser names as keys
         parserInstances = {}
         for parserClass in foundParsers:
-                name = cls.getParserName(parserClass)
+                name = parserClass.getParserName()
                 
                 # Try instantiating the parser
                 try:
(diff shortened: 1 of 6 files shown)


almarklein:
I think this makes sense. The only thing to watch out for is that detecting the Python version from the code should not take long for large files, to avoid long startup times. So only check the first 100 lines or so, perhaps?

ysalmon:
I agree, it would be better to not parse the entire file just for this.
Furthermore, what would be the criterion ?
Possible clues for Python 2 :
- maybe cookie for utf-8 encoding (which if I recall should not be present for Python3 because it is the default) ? 
- ``__future__`` imports about for either ``print_function``, ``division`` or ``unicode_literals``.

This clues, if present, would be in the beginning of a file.

If a file does not have those clues, then either it is Python 3, or a very simple script that can work either in python 2 or python 3 (and we should assume the user wants python 3), or an old python 2 script that is not "python 3 aware" and it should die.

almarklein:
> and it should die.
:'D

I tried to think about some rules, but I fear that whatever Pyzo selects might feel very arbitrary to the user. And, or instance, we'd naturally use python3 for new/empty files, which would suck for python2 users.

So I have another suggestion:
* improve the parser so that we can distinguish keywords from function calls, so we only mark `print` and `exec` as a keyword when its *not* used as a function.
* Color Python2 keywords and builtins in a slightly different (grayer?) color. Not sure how we'd do this. I don't want new entries in the syntax color table, but maybe we can put a modifier (e.g. reduce saturation) on the color to apply ...

ysalmon:
I agree this is tricky. Distinguishing the actual usage of eg. ``print`` as a keyword is not easy because even in Python2 one can write ``print("something")`` in the same way that one can write ``return(0)``. This is bad style imo (and shows a lack of understanding of what a keyword is) but it appears.

Regarding new files : it really should be Python3 by default ; only in existing python2 projects would one want to create Python2 files and one should be aware of this.

Maybe the solution for a useful user experience would be to add a info message at the top of the code editor (like they have in web browsers) with a link/button to switch to the other Python variant ? Is this feasible ?

ysalmon:
So, since there will always be cases where we cannot for sure tell if the code is python2 or python3, it may be better to let the user decide what to do in such a case. I therefore add a ``pythonDefaultDisambiguation`` setting to be used in such a case : it can be set to python2 or python3 for example.

I also added an option to shell configurations to let the user decide which parser to use for each shell. We default to ``defaultStyle``.

ysalmon:
I fumbled with translations and added a new feature to the Workspace. If you want, you can merge everything at once by merging #617 and closing #613 and #616.

almarklein:
FYI I see a lot of issues sn PRs coming in. I am away right now but will have a look at them in august.

almarklein:
Ok, I think that you're right on most points. There's a few tiny things that I want to do differently, but will merge now and apply these in a separate PR.

almarklein:
See https://github.com/pyzo/pyzo/pull/626

PR #614, opened 2019-07-02
Fixing advsettings

ysalmon:
Turns out the Advanced Settings dialog is completely broken ! There is a typo in the code that prevents any actual setting update. Also made help text word-wrap and be translatable, and making the dialog resizable by the user.


almarklein:
Thanks for the catch!

PR #617, opened 2019-07-21
Hide default vars in workspace

ysalmon:
While translating the workspace into French, I thought it would be interesting to give the user (or rather the site administrator) to hide the variables that the IPython shell creates upon startup (``In``, ``Out``, etc.).

I also added a text message to be displayed in place of the variable list when this list is empty.

This PR is on top of #616.


almarklein:
Superseded by #625

almarklein:
(somehow this still shows all the commit from previous pr's, making it hard to review)

ysalmon:
Ah, the PR reference point does not move when master is updated ?

I stopped creating branches on top of each other, restarting from master instead, to avoid too complicated PRs. This has the drawback of forcing a delay in translating, as the translation files are either difficult or impossible to merge in case of a conflict.

almarklein:
Normally it works, I think. But there was a tiny merge conflict with the translation PR. Perhaps that prevented Github from "dropping" existing commits in this PR ...

PR #619, opened 2019-07-21
Fixing despicable bug due to mixup between display string and setting…

ysalmon:
… key


almarklein:
Oh boy, that's one ugly bug that :D
Thanks for the fix!

PR #620, opened 2019-07-21
Two improvements to interactive help

ysalmon:
I tried to set up a back/forward system to let the user navigate between successive queries without retyping them. This is not completely straightforward because there is a query each time we move in the autocompletion list and we obviously do not want these to clutter the history.

There may be some tweaking to decide if (and how) we add a query when autocompletion is validated.
Also, it might be interesting to avoid making queries on very simple terms like ``True``, etc. : if the user types eg. ``numpy.complicated_function(`` he probably wants that function's help to remain displayed, even is some parameter happens to be ``True``. Or we could revert to the earlier query when autocompletion is finished.

I also added an entry in the contextual menu of the editors and shell to query help on the identifier under the mouse cursor (or the selected expression if any). I am using the same logic as the autocompleter to determine what is the identifier under the cursor, ie. ``parseLine_autocomplete``.

Investigating this, I remarked that it will not correctly identify expressions that are not strictly a succession of identifiers, ie. when analysing ``key[1:].lower()`` or ``cursor.block().previous().userState()`` it will not traverse parentheses and therefore not find anything useful. I think this is the same reason why
```
s = "abcd"
s.split()
```
will get autocompletion (or help) when typing the second line, while
```
"abcd".split()
```
will not, although manually querying ```"abcd".split``` works.

``parseLine_signature`` seems more parenthesis-aware.


ysalmon:
Also I think there is [a bug in ``processHelp``](https://github.com/pyzo/pyzo/blob/78c2f57e061a030533242675093447561d7be92c/pyzo/core/baseTextCtrl.py#L400) because ``parseLine_autocomplete`` wants a list of tokens, not a string. So this part of the code will always fail and serve no actual purpose.

almarklein:
Are you happy with the current state of this PR, or does it need more work/discussion?

ysalmon:
I made further improvements : ability to jump in the history + system to catch when help is asked on a keyword.

What bugs me is the interaction between autocompletion and the help system. But maybe this is because autocompletion is not correctly configured on my system  : 
  - I think it autocompletes too early and too much, to the point that it is difficult to type just ``i`` at the end of a line : pressing Enter will complete it to ``id``
 - I am not sure if the selection of an autocompleted entry should go into the help history. Currently it seems not to, I do not remember if I made an effort to attain this.
 - I discovered that the Print utility uses shell.processLine, which will echo the print command before executing it, cluttering the shell display ; can we avoid this ?

ysalmon:
I think this is starting to look good. I managed to tap into the autocompleter to have it restore the previous help entry when it is hidden ; this way it can show help on the possible completions without polluting the history or the help the user explicitly requested.

I also provided help for keywords. I mainly took the official documentation (maybe we should mention that somewhere, for license reasons ?) but rewrote some parts in a more beginner-friendly fashion. Conversely, keywords pertaining to advanced features do not have much help. My idea is not to duplicate the Python docs but to enable beginner to intermediate-level user to be more autonomous during their discovery of the language, so the help messages are mainly written versions of what I have been explaining again and again to my students (even the notion that some words cannot be used as variables because they are reserved keywords is far from easy for some of them).

To make these messages more readable, I added a markdown-like parser for double a triple backquotes (this was to deactivate the treatments made for numpy doc that would mangle lines with ``:`` and to avoid confusion between keywords and the eponymous English words, namely ``for`` and ``and``).

pyzo/codeeditor/extensions/autocompletion.py
```diff
@@ -132,6 +136,8 @@ def autocompleteAccept(self):
     def autocompleteCancel(self):
         self.__completer.popup().hide()
         self.__autocompleteStart = None
+        if self.__cancelCallback is not None :
+            self.__cancelCallback()
```
almarklein: Let's wrap this in a `try - except`:

```py
try:
    self.__cancelCallback()
except Exception as err:
    print("Exception in autocomp cancel callback: " + str(err))
```

ysalmon:
I think we are there.

almarklein:
Awesome, let's merge!

almarklein:
Thanks Yann!

PR #621, opened 2019-07-24
Highlighting non-breakable spaces

ysalmon:
This creates a new token type, IllegalToken, to represent characters that may not appear in Python identifiers (currently, only nbsp will be parsed as such). The default style of this token type is a reddish background.

Consequently, this also fixes some bugs related to background color for characters in editors (the property was not applied !). However, the interaction between styles and the theme editor when adding a new style category remains a mystery for me.

This also fixes an anomalous yet documented behaviour of QtPlainText.toPlainText where nbsp are replaced by regular spaces : this affected us on saving the document.

### DIFF
--- a/pyzo/codeeditor/base.py
+++ b/pyzo/codeeditor/base.py
@@ -225,6 +225,16 @@ class CodeEditorBase(QtWidgets.QPlainTextEdit):
         # Apply a good default style
         self.setStyle(S)
     
+     # see https://bugreports.qt.io/browse/QTBUG-57552?focusedCommentId=469402&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-469402
+     # and https://code.qt.io/cgit/qt/qtbase.git/tree/src/gui/text/qtextdocument.cpp#n1183
+     # and https://doc.qt.io/qt-5/qchar.html
+    _plainTextTrans = str.maketrans({"\u2029":"\n", "\u2028":"\n", "\ufdd0":"\n", "\ufdd1":"\n"})
+    
+    def toPlainText(self) :
+        cursor = self.textCursor()
+        cursor.select(QtGui.QTextCursor.Document)
+        return cursor.selectedText().translate(self._plainTextTrans)
+    
     def _setHighlighter(self, highlighterClass):
         self.__highlighter = highlighterClass(self, self.document())
        
--- a/pyzo/codeeditor/parsers/python_parser.py
+++ b/pyzo/codeeditor/parsers/python_parser.py
@@ -13,7 +13,7 @@ from ..misc import ustr
 from .tokens import (CommentToken, StringToken,
     UnterminatedStringToken, IdentifierToken, NonIdentifierToken,
     KeywordToken, BuiltinsToken, InstanceToken, NumberToken, FunctionNameToken, ClassNameToken,
-    TodoCommentToken, OpenParenToken, CloseParenToken)
+    TodoCommentToken, OpenParenToken, CloseParenToken, IllegalToken)
 
 # Keywords sets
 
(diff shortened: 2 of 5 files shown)


almarklein:
I am having trouble appreciating the use-case :) When is this nbsp a problem?

pyzo/codeeditor/base.py
```diff
@@ -225,6 +225,16 @@ def __init__(self,*args, **kwds):
         # Apply a good default style
         self.setStyle(S)
     
+     # see https://bugreports.qt.io/browse/QTBUG-57552?focusedCommentId=469402&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-469402
+     # and https://code.qt.io/cgit/qt/qtbase.git/tree/src/gui/text/qtextdocument.cpp#n1183
+     # and https://doc.qt.io/qt-5/qchar.html
+    _plainTextTrans = str.maketrans({"\u2029":"\n", "\u2028":"\n", "\ufdd0":"\n", "\ufdd1":"\n"})
```
almarklein: Is this related to our frequent transaltion of `\u2029` to `\n`, e.g. [here](https://github.com/pyzo/pyzo/blob/master/pyzo/core/editorTabs.py#L448)?

ysalmon:
> I am having trouble appreciating the use-case :) When is this nbsp a problem?

I do not know. Apparently, it is a feature : https://bugreports.qt.io/browse/QTBUG-57552

pyzo/codeeditor/base.py
ysalmon: I cannot answer since I do not know the reason for the translation in ``editorTabs.py``; it may be, if QPlainTextEdit uses ``\u2029`` to represent paragraph and this corresponds to a line-break in the code. I just mimicked the behaviour of toPlainText except for nbsp. I since learned that Qt 5.9 introcuced a toRawText method that does not do any replacement.

almarklein:
I mean, in your use-case do you *want* nbsp, or do you want to prevent users from copying nbsp into the code? What problem does this PR solve? :)

ysalmon:
Le 01/08/2019 à 15:29, Almar Klein a écrit :
> I mean, in your use-case do you /want/ nbsp, or do you want to prevent
> users from copying nbsp into the code? What problem does this PR solve? :)

It arose after a discussion with a colleague who is using Pyzo with his
students, and indicated that sometimes they would mistype a nbsp in
their code (fat finger problem imho) which would lead to a
incomprehensible (to them) syntax error : hence the highlighting.

But while testing this we realized that Pyzo would transform nbsp into
spaces upon saving, including nbsp in string literals, which is unwanted
: hence the trick about toPlainText.

ysalmon:
He also complained that with Visible whitespace, spaces are displayed as a median point, which annoys him because he also uses the median point in his texts. But this behaviour from QPlainTextEdit is not configurable yet, in spite of [a feature request at Qt](https://bugreports.qt.io/browse/QTBUG-46072).

ysalmon:
The only thing that is not finished here is fiddling with the themes : there is a new style element and despite my efforts I could not understand why the themes and theme manager behave so weirdly. It seems that some parts were designed upon the hypothesis that all elements and style have the same properties and/or have inconsistent behaviour when a style does not provide values for all properties.

almarklein:
Do you think it's possible to get this in before release? Can you be more specific as to where you need help?

ysalmon:
With the style/theme engine.

The feature in itself is done. But I struggled a lot to get things to more or less work because the interaction between the default style, the syntactic parser, the styling engine, the notion of theme, the theme editor component and the theme description files is very arcane to me.

I added a new stylable element and this seems to breaks things everywhere. I had to "silence" KeyError exceptions because the theme editor was completely crashing.  It seems different parts of the code were written at different times with different assumptions in mind and the whole thing does not "degrade gracefully".

Default styles for tokens are provided in `` pyzo/codeeditor/parsers/tokens.py``. Yet this does not suffice to the theme engine, that has problems when the current theme does not provide a value for the new style element.

almarklein:
I haven't had a lot of time to look into this. I propose that we try to get this into master soon after the release, and do a minor release afterwards.

PR #622, opened 2019-07-24
ShowWhitespace and HighlightMatchingBracket options for shell

ysalmon:
This shuffles options management a little so that ShowWhitespace and HightLightMatchingBracket are also applied to shells, not only code editors.

This adds support for highlighting matching brackets in shell : in fact the heavy work (parsing the line to get tokens) was already done, the tokens were just not stored in the block's userData.


almarklein:
Really nice!

PR #623, opened 2019-07-27
Improve pdf export

filman230:
Hello!
I improved the pdf export system, which was very basic.

Now, there are some options, such as page orientation, enable/disable syntax highlighting, and lines numbers have now a grey background. There is also a preview.

pdfExport

I tried to add options for margins, but unfortunately it seems there is a bug with Qt:
https://stackoverflow.com/questions/49351391/qprinter-margins-under-py3-and-pyqt5
https://bugreports.qt.io/browse/QTBUG-32987

Also, I wanted to add an option for page number, but there is no simple way to change this (AFAIK).

This is probably far from perfect, so please tell me if there is any problem (bad function/variable name, or anything else).

### DIFF
--- a/pyzo/core/menu.py
+++ b/pyzo/core/menu.py
@@ -28,6 +28,8 @@ from pyzo.core.pyzoLogging import print  # noqa
 from pyzo.core.assistant import PyzoAssistant
 from pyzo import translate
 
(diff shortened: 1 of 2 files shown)


almarklein:
Nice!  This is a real improvement :D

PR #628, opened 2019-08-06
respecting option for removing trailing whitespace on save

Spaak:
Previously, the editor would always strip trailing whitespace on save, even when the setting is set to 0, because the setting is compared using `or` rather than `and`. This PR fixes this.

### DIFF
--- a/pyzo/core/editor.py
+++ b/pyzo/core/editor.py
@@ -499,7 +499,7 @@ class PyzoEditor(BaseTextCtrl):
             return
 
         # Remove whitespace in a single undo-able action
-        if self.removeTrailingWS or pyzo.config.settings.removeTrailingWhitespaceWhenSaving:
+        if self.removeTrailingWS and pyzo.config.settings.removeTrailingWhitespaceWhenSaving:
             # Original cursor to put state back at the end
             oricursor = self.textCursor()
             # Screen cursor to select document


almarklein:
The code is actually correct, but I understand that it may be a bit confusing. The reason is historical: first, Pyzo never removed trailing ws. Then, we added the `removeTrailingWhitespaceWhenSaving` setting so users could turn it on. And recently, we made Pyzo remove trailing ws when the file already contained very little trailing ws when it was loaded.

Spaak:
Thanks for the reply; I understand the `or` now. Still, it's a bit annoying that Pyzo now removes whitespace on certain files (apparently when it contains little trailing ws), with no way of disabling this.

> And recently, we made Pyzo remove trailing ws when the file already contained very little trailing ws when it was loaded.

I don't really understand why that would be useful in general: little whitespace -> remove irrespective of preference; lots of whitespace -> look at preference. I'd rather have Pyzo _never_ remove any trailing ws, for which there now appears to be no setting. Or in other words, why would the amount of whitespace matter for (1) whether or not to remove it; and more importantly, for (2) whether or not to respect a user-defined preference?

(My hunch is by the way that for Python, not removing ws is a widespread preference, since this messes up the indentation guides. It only really might be preferable to remove trailing ws if users have indentation guides switched off.)

almarklein:
> My hunch is by the way that for Python, not removing ws is a widespread preference, since this messes up the indentation guides.

This has been my hunch too, but over the years I've seen that more and more editors remove whitespace by default. And to be honest, the indentation guides is not a really good argument - editors can still draw them.

So maybe your pain would be resolved if Pyzo draws indentation guides over empty lines where it makes sense?

> I don't really understand why that would be useful in general: little whitespace -> remove irrespective of preference; lots of whitespace -> look at preference

The motivation is that if you have a file that already has no (or barely any) whitespace, Pyzo assumes that it can safely remove all whitespace. When there is more whitespace, Pyzo keeps it as-is (I don't like these PR's that change 3 lines, plus 100 more because an editor removed the ws).

Issue #627, opened 2019-08-02
Unify and reduce margins


kelsa-pi:
Margins between gui elements editor, shell, logger and plugins are not consistent  as shown in the image below. Second, they are to thick IMHO. 
My suggestion is to unify and reduce margins. It would increase screen real estate and improve look & feel (for me anyway). I guess this can be done by config option/s.

Image description:
Left - current margins, setContentsMargins(0,0,0,0) and setContentsMargins(4,4,4,4), for more details [pyzo search](https://github.com/pyzo/pyzo/search?q=setContentsMargins&unscoped_q=setContentsMargins)
Right - very exaggerated example, unfied and reduced margins for editor, shell, logger and plugins ( setContentsMargins(0,0,0,0) )

![margins](https://user-images.githubusercontent.com/19865053/62395617-1e8d0100-b571-11e9-894f-7705b1c53a9f.png)

almarklein:
Making the margins more consistent would certainly be a good idea. I'm not sure if I'd want them to be much smaller; it does give a bit of extra space, but it also makes it feel more "cramped".

almarklein:
Fixed by #629

PR #629, opened 2019-08-06
Unify margins between gui elements editor, shell and plugins #627

kelsa-pi:
Margins between gui elements editor, shell, logger and plugins are consistent as shown in the image below. New config option added: margins=[4,4,4,4].

new_margins

### DIFF
--- a/pyzo/tools/pyzoLogger.py
+++ b/pyzo/tools/pyzoLogger.py
@@ -34,8 +34,8 @@ class PyzoLogger(QtWidgets.QWidget):
         # spacing of widgets
         self.layout.setSpacing(0)
         # set margins
-        margins = pyzo.config.view.widgetMargins
-        self.layout.setContentsMargins(*margins)
+        margin = pyzo.config.view.widgetMargin
+        self.layout.setContentsMargins(margin, margin, margin, margin)
         self.setLayout(self.layout)
 
     def updateZoom(self):
--- a/pyzo/tools/pyzoLogger.py
+++ b/pyzo/tools/pyzoLogger.py
@@ -20,23 +20,30 @@ class PyzoLogger(QtWidgets.QWidget):
 
         The main widget for this tool.
 
-        """
-    
+    """
+
     def __init__(self, parent):
         QtWidgets.QWidget.__init__(self, parent)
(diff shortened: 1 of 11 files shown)


pyzo/tools/pyzoLogger.py
```diff
@@ -6,17 +6,40 @@
...
+from pyzo.util.qt import QtCore, QtGui, QtWidgets  # noqa
 from pyzo.core.shell import BaseShell
 from pyzo.core.pyzoLogging import splitConsole
 
-
 tool_name = pyzo.translate("pyzoLogger","Logger")
 tool_summary = "Logs messages, warnings and errors within Pyzo."
+
+
+class PyzoLogger(QtWidgets.QWidget):
```
almarklein: This wrapper class will cause problems, because in some places we use `pyzo.toolManager.getTool('pyzologger')` and do stuff with it. It it possible to fix the margins without wrapping the widget in another widget? Otherwise, this class will need a public property to get access to the shell, and all occurrences have to be updated.

almarklein:
I like this a lot!

> In config: `margins = [4,4,4,4]`

Is there a use-case where a user would not want uniform margins? I'd prefer doing `margins = 4` (or "widget_margins"? your call) in the config, and `xx.setContentsMargins(margin, margin, margin, margin)` in the code.

kelsa-pi:
In pyzoLogger class added: updateFont() and updateZoom().

In config margins = [4,4,4,4] replaced by widgetMargins = [4,4,4,4]

kelsa-pi:
>

almarklein:
> In pyzoLogger class added: updateFont() and updateZoom().

Great, that should do the trick.

> In config margins = [4,4,4,4] replaced by widgetMargins = [4,4,4,4]

My comment was mostly about replacing `[4,4,4,4]` with just `4`. That would make things less fragile (e.g. if a user changes the config and does not put exactly 4 numbers in there).

kelsa-pi:
>  That would make things less fragile (e.g. if a user changes the config and does not put exactly 4 numbers in there).

+1

almarklein:
Thanks a lot!

PR #631, opened 2019-08-18
Use standard folders for data and config

filman230:
On Unix-like systems, pyzo config folder is located directly inside home folder of users, which is a bit annoying: https://0x46.net/thoughts/2019/02/01/dotfile-madness/

Pyzo should use $XDG_CONFIG_HOME to store configuration, and $XDG_DATA_HOME to store data (e.g. history). (https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html)
I used QStandardPaths from Qt to get those paths, this should work on any system (MacOs, Windows...).
For retro-compatibility reasons, pyzo will use old its folder if it exists.

I don't really understand the purpose of pyzo_dirs2() and pyzo_dirs(). What path should they use ?

### DIFF
--- a/pyzo/util/paths.py
+++ b/pyzo/util/paths.py
@@ -20,6 +20,7 @@ This code was first part of pyzolib, and later moved to pyzo.
 # * See docstring: that's why the functions tend to not re-use each-other
 
 import sys
+from pyzo.util.qt import QtCore
 
 ISWIN = sys.platform.startswith('win')
 ISMAC = sys.platform.startswith('darwin')
@@ -72,7 +73,7 @@ def temp_dir(appname=None, nospaces=False):
         path = os.path.join(path, appname)
         if not os.path.isdir(path):
             os.mkdir(path)
-    
+
     # Done
     return path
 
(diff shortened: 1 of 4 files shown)


pyzo/util/paths.py
```diff
@@ -130,18 +136,36 @@ def appdata_dir(appname=None, roaming=False, macAsLinux=False):
...
-    # Done
-    return path
+            # Better way to get config/data folder especially on *nix system (see XDG_CONFIG_HOME standard),
+            # but this should work on any os.
+            # For retro-compatibility, check if old folder exist, and if not, use standard path.
+            standard_data_path = QtCore.QStandardPaths.writableLocation(QtCore.QStandardPaths.AppDataLocation)
+            standard_config_path = QtCore.QStandardPaths.writableLocation(QtCore.QStandardPaths.ConfigLocation)
+
+            # Check if QStandardPaths succeeded to find the location, otherwise use old path
+            if standard_config_path != "" and standard_data_path != "":
```
almarklein: Maybe simply `if not (standard_config_path and standard_data_path):`, that also feels a bit safer  in case Qt ever returns `None`.

almarklein:
> I don't really understand the purpose of pyzo_dirs2() and pyzo_dirs(). What path should they use ?

You can probably ignore these. That's from when Pyzo is from back when Python was a distribution + IDE.

almarklein:
Ah, let's just merge this now

almarklein:
Thanks!

almarklein:
Tweaked my suggestion after merge.

PR #637, opened 2019-08-27
Status bar - add cursor position and file encodning info

kelsa-pi:


pyzo/core/main.py
```diff
@@ -185,13 +186,11 @@ def _populate(self):
...
-        if pyzo.config.view.showStatusbar:
-            pyzo.status = self.statusBar()
-        else:
-            pyzo.status = None
-            self.setStatusBar(None)
-        
+        # Create status bar
+        self.setStatusBar(StatusBar())
+        # show or hide
+        self.statusBar().setVisible(pyzo.config.view.showStatusbar)
```
almarklein: I was going to say that we need an entry in the default config, but then I saw we already have it. haha

almarklein:
Looking good, merging, thanks!

almarklein:
@kelsa-pi I found a bug: the shown column is the column in the editor, not the position on the line. This becomes an error when a block is wrapped over multiple lines.

almarklein:
Fixed in #646

PR #640, opened 2019-09-04
let QMenu display its tooltips

### LINKED ISSUE #638: Black artifact windows triggered by menu tooltips
On Windows 7 (when using a 64bit Pyzo with PyQt5) annoying black rectangular artifacts pop up 
when hovering menu items providing a tooltip. Modal windows that stay on top system-wide. Almost invisible when they manage to collapse to a small black line -- but sometimes they occupy a square of about 190x160 pixels.

Inspired by #399 I tweaked the environment variable `QT_GRAPHICSSYSTEM`. But to no avail. 

Hence I suspect this issue isn't related to the graphics mode but to the way how tooltips for menu items are created.

Only workarounds so far:

- Use keyboard to select a menu item.
- Slow down when clicking/hovering menu items. Let the artifacts flicker until they collapse to a minuscule black line.

Tontyna:
should fix #638

Bonus: with Qt >= 5.1 the context menus display their tooltips, too

### DIFF
--- a/pyzo/core/menu.py
+++ b/pyzo/core/menu.py
@@ -73,7 +73,8 @@ def buildMenus(menuBar):
         if hasattr(action, '_shortcutsText'):
             tt = tt + ' ({})'.format(action._shortcutsText) # Add shortcuts text in it
         QtWidgets.QToolTip.showText(QtGui.QCursor.pos(), tt)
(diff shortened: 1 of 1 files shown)


almarklein:
Oh, this is great. Thanks for looking into it!

PR #648, opened 2019-09-28
Overwrite an existing shell settings with backup shell settings.

kelsa-pi:
After Factory Defaults, it would be nice to have option to overwrite an existing (Python) shell settings with backup shell settings.


almarklein:
Thanks!

PR #651, opened 2019-10-06
Portable settings folder

Tontyna:
Renaming `_settings` folder to `settings` has no *portable* effect if there isn't a `pyzo` subfolder.


almarklein:
Nice catch, thanks!

PR #674, opened 2020-01-29
Fix windows installer and retina display support

almarklein:


almarklein:
This fixes the 64bit installer, but the 32bit windows installer produces a broken Pyzo on my system. Ah well, progress, I guess :)

PR #679, opened 2020-02-03
Build tweaks

### LINKED ISSUE #678: pyzo v4.10.0 binary crashes on Linux
OMG! Extracting `pyzo-4.10.0-linux64.tar.gz` and starting the `pyzo` launcher from terminal pyzo crashes with `Segmentation fault` as soon as I press a key on my keyboard.

The terminal output indicates besides a lot of `Gtk-WARNING: **: Theme parsing error:...` that pyzo's FontConfig has been updated since v4.9.0 -- `Unable to update the static FcBlanks: 0x0600` -- and finally `Segmentation fault`.

System is Ubuntu Bionic
Machine: x86_64
Version: <span>#</span>40~18.04.1-Ubuntu SMP Thu Nov 14 12:06:39 UTC 2019
Release: 5.0.0-37-generic

Doesn't help to exchange the `source/pyzo` with current master-branch source, including #676.

----

Summary of this issue (by AK):

Observations:
* There are some warning/error messages, but Pyzo startsup just fine. It segfaults the moment a key is pressed.
* The build is performed on Azure Pipelines, on an Ubuntu 16.04 VM.
* Somehow the resulting frozen app contains a different set of libs than Pyzo v4.9.0 which was also build on Ubuntu 16.04 (but via a Virtualbox VM). It mostly contains *more* libs, but some libs are also "missing".
* `libxkbcommon` seems to be causing the segfault.
* The `v4.9.0` build has `libxkbcommon.so.0` and `libxkbcommon-x11.so.0` while in v4.10.0 there is only the former. If I either remove `libxkbcommon`, or add `libxkbcommon-x11`, the segfault is gone.

Questions:
* Why is PyInstaller incapable of selecting the correct set of dependencies in this situation?
* What can we do to fix it?

Things I tried (and did not work unless mentioned otherwise):
* Use older versions of PyInstaller, and also the bleeding edge dev version.
* Different Python versions via the `UsePythonVersion` Azure Pipelines task.
* Using a ppa to install Python 3.6 (Pyzo needs 3.6, while 16.04 has 3.5). 
* Using Docker to build on a ManyLinux VM. Ha! Not suitable for PyInstaller because it (deliberately does not contain libpython).
* PySide2 instead of PyQt5.
* Use 18.04: running the resulting application does not popup Pyzo but simply freezes.
* Using Docker `python:3.7-stretch`: works on 18.04, and no fontconfig warnings :). On 20.04 there are many fontconfig warnings and it works!

almarklein:
* Only build windows installer for 64bit (because the 32bit version seems to be broken).
* Make that 64bit installer install in `program files`, not `program file x86`.
* Try to fix #678 ... no luck so far. Ideas:
  * build the Linux version on Travis? (and then figure out how to push the artifact *sigh*)
  * build on another vm?
  * use the default python on the vm instead of using the `UsePythonVersion` task.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -26,6 +26,7 @@ jobs:
         vmImage: ubuntu-16.04
         TEST: 1
         FREEZE: 1
+        PY_VERSION: 3.5
       MacOS-64:
         vmImage: macOS-10.14
         MACOSX_DEPLOYMENT_TARGET: '10.13'
@@ -58,7 +59,7 @@ jobs:
         set -ex
         python -m pip install -U pip
         pip install -U pyqt5
-        pip install -U pytest black flake8 pyinstaller==3.5
+        pip install -U pytest black flake8 pyinstaller==3.6
   - task: Bash@3
     displayName: Install inno setup (on Windows)
     condition: eq(variables['Agent.OS'], 'Windows')
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -97,7 +97,7 @@ jobs:
       script: |
         set -ex
         $(PY_EXE) setup.py sdist
-        pip install $(echo "$(Build.SourcesDirectory)" | sed -e 's/\\/\//g')/dist/*.tar.gz
+        $(PY_EXE) -m pip install $(echo "$(Build.SourcesDirectory)" | sed -e 's/\\/\//g')/dist/*.tar.gz
   - task: Bash@3
     displayName: Test
     condition: and(succeeded(), eq(variables['TEST'], '1'))
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -10,7 +10,7 @@ variables:
   NAME: pyzo
   PY_ARCH: x64
   PY_VERSION: 3.7
-
+  PY_EXE: python
 
 jobs:
 
@@ -26,6 +26,8 @@ jobs:
         vmImage: ubuntu-16.04
         TEST: 1
         FREEZE: 1
+        PY_VERSION: native
+        PY_EXE: python3
       MacOS-64:
         vmImage: macOS-10.14
         MACOSX_DEPLOYMENT_TARGET: '10.13'
@@ -47,6 +49,7 @@ jobs:
   steps:
   - task: UsePythonVersion@0
     displayName: Select Python
+    condition: ne(variables['PY_VERSION'], 'native')
     inputs:
       versionSpec: $(PY_VERSION)
       architecture: $(PY_ARCH)
@@ -56,9 +59,9 @@ jobs:
       targetType: inline
       script: |
         set -ex
-        python -m pip install -U pip
-        pip install -U pyqt5
-        pip install -U pytest black flake8 pyinstaller==3.6
+        $(PY_EXE) -m pip install -U pip
+        $(PY_EXE) -m pip install -U pyqt5
+        $(PY_EXE) -m pip install -U pytest black flake8 pyinstaller==3.6
   - task: Bash@3
     displayName: Install inno setup (on Windows)
     condition: eq(variables['Agent.OS'], 'Windows')
@@ -83,7 +86,7 @@ jobs:
       targetType: inline
       script: |
         set -ex
-        python setup.py sdist
+        $(PY_EXE) setup.py sdist
         pip install $(echo "$(Build.SourcesDirectory)" | sed -e 's/\\/\//g')/dist/*.tar.gz
   - task: Bash@3
     displayName: Test
@@ -92,7 +95,7 @@ jobs:
       targetType: inline
       script: |
         set -ex
-        pytest --junit-xml=results.xml tests || true
+        $(PY_EXE) -m pytest --junit-xml=results.xml tests || true
         # Fail the task if results.xml was not created
         if [[ ! -f results.xml ]]
         then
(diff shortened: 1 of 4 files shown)


almarklein:
After spending quite some time trying to freeze on manylinux, I learned that manylinux does not have libpython, so pyinstaller can't do it's thing: https://github.com/pypa/manylinux/issues/255

Tontyna:
I'm almost certain that the segfault is a result of mismatched libraries -- it's a matter of coincidence when a Linux binary happens to work. BTW: On what system did you build the v4.9.0 tarball?

Won't help to use aged versions of Python/PyInstaller/Qt. It's the system libraries not matching the frozen libraries.

Quoting the [pyinstaller FAQ](https://github.com/pyinstaller/pyinstaller/wiki/FAQ#gnulinux):

> The executable that PyInstaller builds is **not fully static**, in that it still depends on the system libc. [..] you can't run the resulting executable on an older system. 

Though they are talking about GLIBC, I suspect that other libraries are affected, too (remember [your pyzo](https://github.com/pyzo/pyzo/issues/678#issuecomment-581478959)  `Failed to load module: /usr/lib/x86_64-linux-gnu/gio/modules/libgiognutls.so`)

The FAQ suggest to either link against *the oldest library version* available. Or:

> Another solution is to use a tool like [StaticX](https://github.com/JonathonReinhart/staticx/) to create a fully-static bundled version of your PyInstaller application.

almarklein:
Yeah, but I figured that a different PyInstaller version might ship the correct libs. Pyzo v1.9 was build on a Ubuntu 16.04 VM. That's also why I am so surprised (and frutstrated) that it does not work. This does not seem like a very special situation; just a Qt app build on a very standard Linux. My hope is now at using the Azure VM's default Python. But I have other things to do right now. Will try later today/this week.

almarklein:
Linux build now works on 16.04 with python3.6 installed via a ppa. The system should now be very similar to the VM that I've been using to build v4.9.0. Alas, same segfault :'(

Tontyna:
> Alas, same segfault

It's difficult but possible to work with pyzo without hitting any key on the keyboard :grimacing:

Tontyna:
That's written into my `/var/log/kern.log/` when I hit a key in frozen pyzo:

```
Feb  4 23:33:34 MISOX kernel: [ 3150.987797] pyzo[4003]: segfault at a ip 00007f81a226e688 sp 00007ffe7211a030 error 4 in libxkbcommon.so.0[7f81a2252000+3d000]
Feb  4 23:33:34 MISOX kernel: [ 3150.987811] Code: 00 00 89 d2 48 8d 04 52 48 8b 44 c6 08 8b 50 18 23 78 08 85 d2 74 2e 48 8b 40 20 83 ea 01 48 8d 54 92 05 48 8d 0c 90 0f 1f 00 <8b> 50 04 85 d2 8b 50 08 74 04 85 d2 74 04 39 d7 74 0b 48 83 c0 14
```

Does it help to know that it's `error 4 in libxkbcommon.so.0`?

almarklein:
> Does it help to know that it's error 4 in libxkbcommon.so.0?

Ha, I don't even know what `libxkbcommon` is for ...

*does internet search:*

> xkbcommon is a library for handling of keyboard descriptions, including loading them from disk, parsing them and handling their state. It's mainly meant for client toolkits, window systems, and other system applications; `

almarklein:
So in v4.9.0 there is `libxkbcommon.so.0` and `libxkbcommon-x11.so.0` while in v4.10.0 there is only the former. If I either *remove* libxkbcommon, or *add* libxkbcommon-x11, the segfault is gone.

So this could be a fix. But it feels fragile, and there are still multiple warnings ...

Tontyna:
Assembling the files required by the frozen app is the job of PyInstaller, we shouldn't be forced to re-work that.
AFAIK collecting the **so**-files is delegated to `gi.repository` which does its job via *gobject-introspection* which asks the current system for the GObjects in use, their dependencies and the corresponding  libraries.

Does the **Platform: Linux-4.15.0-1066-azure-x86_64-with-Ubuntu-16.04-xenial** have a tuned package system, libraries optimized/enhanced for running in the Azure Linux VM? Not suitable for running the frozen Python app outside their VM? 

When I run `freezeScript.py` from current master branch on my machine **Platform: Linux-5.0.0-37-generic-x86_64-with-Ubuntu-18.04-bionic** the produced binary runs as expected, no segfault.

Dunno whether the azure build would be fine if I had a kernel major release number of **4** ?

almarklein:
> Does the Platform: Linux-4.15.0-1066-azure-x86_64-with-Ubuntu-16.04-xenial have a tuned package system [...]


It looks like *something* is different, causing pyinstaller to not do its job correctly.

almarklein:
Tried to build on Docker `python:3.7-stretch`: works on 18.04, and no fontconfig warnings :). You can try the `4.10.0.7` release.

Unfortunately, on 20.04 there are many fontconfig warnings and it freezes before showing a gui :(

I am tempted to guide Linux users to make sure they have Python3.6 and Pyqt5 or PySide2, and then pip install pyzo ...

Tontyna:
Yepp, 4.10.0.7 works for me.

Though it's not too difficult to build your own frozen binary, it's easier to tell people how to run a pip installed pyzo.

almarklein:
> Unfortunately, on 20.04 there are many fontconfig warnings and it freezes before showing a gui :(

I just opened my VM and there were several Pyzo's open. I guess it just took some time to load the first time for some reason. It seems to work fine! :D:D

Also tried to use Docker using image `ubuntu:16.04`, which took a lot of tweaking (see commits ;) but did not work at all: `Could not load the Qt platform plugin "xcb" in "" even though it was found.`

almarklein:
Better squash these commits ...

PR #681, opened 2020-02-07
Displaying signatures in interactive help

ysalmon:
Contrary to the shell's ``help`` builtin, Pyzo's interactive help will not display the signatures of functions.
This is due to :

- the interactive help not doing anything with the signature when it gets one ;
- the introspection module erasing the signature from its response on builtin functions.

According to the comments, the latter is done on the assumption that builtin functions have their signature at the beginning of their docstring. This is true for eg. ``numpy.zeros`` but false for ``open``.

There is some code in the help tool to detect the signature in the docstring and displaying it nicely ; I added a fallback code where we just use the signature given by the introspection.

### DIFF
--- a/pyzo/pyzokernel/introspection.py
+++ b/pyzo/pyzokernel/introspection.py
@@ -392,8 +392,9 @@ class PyzoIntrospector(yoton.RepChannel):
 
             # get and correct signature
             h_fun, kind = self._getSignature(objectName)
-            if kind == "builtin" or not h_fun:
-                h_fun = ""  # signature already in docstring or not available
+
+            if not h_fun:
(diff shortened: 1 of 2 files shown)


almarklein:
Thanks! As you can see we have CI now! :)  It only checks some formatting and flake8, and checks if import pyzo works. Over time we can add more actual tests. The most important thing is that it also builds the binaries, making it *a lot* easier to do a release.

For this PR this means that you need to run `black .` in the root of the repo, to apply the auto-formatting, and maybe a `flake8 .` to check for flake errors.

almarklein:
Thanks Yann!

PR #687, opened 2020-02-13
Try fixing Linux binaries ... again

### LINKED ISSUE #686: fozen Linux Pyzo: "File|Open" crashes
Binary Pyzo v4.10.2 on Linux crashes when I try to open a file. Console says:

```
symbol lookup error: /usr/lib/x86_64-linux-gnu/libgtk-3.so.0: undefined symbol: g_log_structured_standard
```

Seems to be another case of library mismatch :anger: :sob:

almarklein:
Closes #686

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -34,7 +34,7 @@ jobs:
         TEST: 0
         FREEZE: 0
         #PY_EXE: python
-        DOCKERIMAGE: ubuntu:18.04
+        DOCKERIMAGE: ubuntu:16.04
       Build-MacOS-64:
         vmImage: macOS-10.13
         MACOSX_DEPLOYMENT_TARGET: '10.13'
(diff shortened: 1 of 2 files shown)


Tontyna:
This fixes the crashing "File|Open" -- at least on (my) Ubuntu 18.04

The price to pay is `Failed to load module "canberra-gtk-module"`.
This is just a Gtk-*Message*. No problem. Don't need no event sounds.

@almarklein It's impossible to '*fix Linux binaries*'. Not when freezing a Qt that obviously depends on dynamically loaded Gtk libraries and (of course) doesn't put the whole Gtk into the frozen binary's `LD_LIBRARY_PATH` but only a basic collection -- in the hope that the omitted files will be present on the target system, in the hope the omitted files won't mismatch/segfault/...

I mean: The `pyzo-4.10.2.2-linux64.tar.gz` was built with ubuntu-18.04 and Python 3.6.9. Same Ubuntu version here, same system Python here -- but nevertheless: library mismatch...

almarklein:
yeah ... it looks like it's at least *more* compatible ;)

PR #690, opened 2020-02-18
Starting autocomplete after some chars

### LINKED ISSUE #689: Do not start autocompletion from first character
When I am typing a short variable like ``i`` in a loop, Pyzo pops up its autocompletion list with everything that begins with ``i``, which is more annoying than helpful.

I think there was an option to start autocompletion after only 3 or 4 characters at a point. Or I am confusing with another IDE ?

In any case I propose to add it (back ?).

ysalmon:
Fixes #689 
Currently showing the popup after 3 chars. Seems to work.

**How to manage the setting ?** Should AutoCompletion refer directly to the value of pyzo setting, or should make the setting separate from AutoCompletion::__autocompleteMinChars ? This would require an update mechanism to propagate a new setting value.

Currently the interactive help still updates from the first char, which is odd. **Should the interactive help also "listen" to the pyzo setting ?** I think it would be to mundane to make a setting for the number of chars before the popup and another for updating the interactive help.

### DIFF
--- a/pyzo/codeeditor/extensions/autocompletion.py
+++ b/pyzo/codeeditor/extensions/autocompletion.py
@@ -47,6 +47,10 @@ class AutoCompletion(object):
         # Text position corresponding to first charcter of the word being completed
         self.__autocompleteStart = None
 
+        # We show the popup when this many chars have been input
+        self.__autocompleteMinChars = 3
+        self.__autocompleteVisible = False
+
         self.__autocompleteDebug = False
 
         self.__autocompletionAcceptKeys = (Qt.Key_Tab,)
@@ -117,6 +121,7 @@ class AutoCompletion(object):
         if (
             not self.autocompleteActive()
             or startcursor.position() != self.__autocompleteStart.position()
+            or not self.autocompleteVisible()
         ):
 
             self.__autocompleteStart = startcursor
@@ -125,7 +130,8 @@ class AutoCompletion(object):
             # Popup the autocompleter. Don't use .complete() since we want to
             # position the popup manually
             self.__positionAutocompleter()
-            if self.__updateAutocompleterPrefix():
+            if self.__updateAutocompleterPrefix() and len (self.__completer.completionPrefix()) >= self.__autocompleteMinChars:
+                self.__autocompleteVisible = True
                 self.__completer.popup().show()
             if self.__autocompleteDebug:
                 print("self.__completer.popup().show() called")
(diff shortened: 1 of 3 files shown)


almarklein:
>  Should AutoCompletion refer directly to the value of pyzo setting, or should make the setting separate from AutoCompletion::__autocompleteMinChars ? 

Not sure if I understand you correctly :) Do you mean whether we should make it configurable via the config? I think that makes sense. Adding it to the `defaultConfig.ssdf` will make sure that the value is created when missing.

> Should the interactive help also "listen" to the pyzo setting ? 

Yes, please let's use one and the same value for both ...

ysalmon:
Le 24/02/2020 à 11:12, Almar Klein a écrit :
>     Should AutoCompletion refer directly to the value of pyzo setting,
>     or should make the setting separate from
>     AutoCompletion::__autocompleteMinChars ?
> 
> Not sure if I understand you correctly :) Do you mean whether we should
> make it configurable via the config? 

Yes, of course.

My question is how.

We can either

- have the code of the autocomplete explicitely depend on pyzo settings
: each time it tries to perform an autocomplete, it reads some
pyzo.settings.whatever

- or maintain its genericity by having a setAutocompleteMinChars method
that would be independant of the rest of pyzo's modules ; the settings
manager would then have to call this method for all instances of the
autocompleter when the corresponding setting is changed.

almarklein:
The second option. This is how it works for all editor-related config. The editor component does not know about Pyzo.

ysalmon:
OK.

Is it the same for the interactive help ? It imports pyzo.

almarklein:
No, the tools can use Pyzo's internal API :)

ysalmon:
In fact I now catch too-short-prefixes earlier in the autocompleter, so
they do not leak to interactive help at all : nothing to do there.

I added a setting in defaultConfig.ssdf and a method
setAutocompleteMinChars to the AutoCompletion class.

How do we call this method on setting change ?

almarklein:
Nice work! That's probably `core/baseTextCtrl.py`, in the `__init__` of `class BaseTextCtrl`. Or in `shell.py` if it only applies to shells, or `editor.py` if it only applies to editors.

ysalmon:
OK, so we apply the setting at startup. It means a restart is required when it is changed, but I think that is good enough.

almarklein:
LGTM, ready from your end?

ysalmon:
Yes.

almarklein:
Thanks Yan!

PR #692, opened 2020-02-27
Making autodetection more automatic

### LINKED ISSUE #688: Improving auto-detection of Python
Since Pyzo has ceased to be an IDE/interpreter bundle, we have an autodetection mechanism for the python interpreter at first launch. This systems works quite well but asks the user for a confirmation.

To eschew this step, I contributed the siteConfig.ssdf mechanism which allows inter alia a site admin to preset the shell configuration for new users. This works quite well if the admin installs both pyzo and the related interpreter in know paths, or just uses /usr/bin/python3.

But it came to light that people using this with arrays of Windows computers with non-uniform python installation are facing the following problem : the python interpreter tends to be installed in a path that depends on its version. Hardcoding the path in siteConfig.ssdf becomes a problem more than a solution.

I therefore propose to add a setting that would allow a site admin (via siteConfig.ssdf) to let the autodetection mechanism skip the user confirmation.

But then the chosen interpreter path would be hardcoded in the user's settings, which could cause problems when the site admin replaces an old python with a new one, or the users logs into another computer where the python version is different.

To avoid this I propose to tweak the interpreter path property of shells to make it possible to explicitly say that the interpreter should be re-detected each time. It should be possible to store as path a string like ``:/:/:autodetect`` which I think is not a valid path on any filesystem.

ysalmon:
Fixes #688 

This adds a new setting auto_useFound :
- if 0, the current behaviour is retained,
- if 1, upon autodetecting a python interpreter, the confirmation step will be skipped, 
- if 2, autodetection will be performed even if there already exists a shell in the configuration.

To avoid spamming the shell config list with new shells in mode 2, when a shell config is created from autodetection, it is marked as such (wasAutodetected property set to 1). The useFound method will scan the existing shell configs to find a config with this property, and will update the exe in that config.

Incidentally, this sets the "use IPython if available" property to 1 by default for new shell configs.

This also changes the behaviour of pyzo when the starting config does not have a shellConfigs2 at all : instead of populating it with a spurious shell config with the empty string as exe, it creates an empty list. The previous behaviour prevented autodetection from working in such a case.

EDIT : this now works. Instead of calling useFound from detect, I set a 100 ms QTimer to do it. This seems to let everything that's needed be properly created.

The PR works as intended. In mode 2 in particular, additional shell configs that were manually created by the user remain untouched.

To avoid a problem where the user is in mode 2 but has modified the autodetected exe and it gets overwritten at next start, the shell config editor is tweaked to remove the wasAutodetected flag when a config exe is modified.

The shell config that gets started at the beginning is the config that was just created/updated with useFound, if any, or the first one in the config list, if useFound was not called.

### DIFF
--- a/pyzo/core/kernelbroker.py
+++ b/pyzo/core/kernelbroker.py
@@ -74,7 +74,7 @@ class KernelInfo(ssdf.Struct):
         self.gui = "Auto"
 
         # Use IPython if available
-        self.ipython = 'yes'
+        self.ipython = "yes"
 
         # The Python path. Paths should be separated by newlines.
         # '$PYTHONPATH' is replaced by environment variable by broker
--- a/pyzo/core/kernelbroker.py
+++ b/pyzo/core/kernelbroker.py
@@ -73,6 +73,9 @@ class KernelInfo(ssdf.Struct):
         # Instantiate with a value that is settable
         self.gui = "Auto"
 
+        # Use IPython if available
+        self.ipython = 'yes'
+
         # The Python path. Paths should be separated by newlines.
         # '$PYTHONPATH' is replaced by environment variable by broker
         self.pythonPath = ""
--- a/pyzo/core/shellStack.py
+++ b/pyzo/core/shellStack.py
@@ -622,13 +622,13 @@ class InterpreterHelper(QtWidgets.QWidget):
         self.setLayout(layout)
         layout.addWidget(self._label, 1)
 
-    def refresh(self, noRedetect = False):
+    def refresh(self):
         self._label.setText("Detecting interpreters ...")
         QtWidgets.qApp.flush()
         QtWidgets.qApp.processEvents()
-        self.detect(noRedetect)
+        self.detect()
 
-    def detect(self, noRedetect = False):
+    def detect(self):
 
         python_link = '<a href="https://www.python.org/">Python</a>'
         conda_link = '<a href="https://miniconda.pyzo.org">Miniconda</a>'
@@ -636,9 +636,8 @@ class InterpreterHelper(QtWidgets.QWidget):
         configs = pyzo.config.shellConfigs2
 
         # Hide now?
-        if (noRedetect or pyzo.config.settings.auto_useFound < 2) and configs and configs[0].exe:
-            self._label.setText("Happy coding!")
-            QtCore.QTimer.singleShot(1200, self.hide_this)
+        if pyzo.config.settings.auto_useFound < 2 and configs and configs[0].exe:
+            self.finish()
             return
 
         # Always sleep for a bit, so show that we've refreshed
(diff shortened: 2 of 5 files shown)


ysalmon:
Huh, the Linux build check failed because docker failed to install some dependencies…

pyzo/core/shellStack.py
```diff
@@ -725,21 +726,49 @@ def useFound(self):
...
+            if pyzo.config.settings.auto_useFound == 2:
+                for conf in configs:
+                    if getattr(conf, "wasAutodetected", False):
+                        config_to_use = conf
+                        break
+            if (
+                len(configs) == 0
+                or pyzo.config.settings.auto_useFound == 2
+                and config_to_use is None
+            ):
```
almarklein: This can be `if config_to_use is None`, right? (and then the `if` below can be an `else`)

pyzo/core/kernelbroker.py
```diff
@@ -73,6 +73,9 @@ def __init__(self, info=None):
         # Instantiate with a value that is settable
         self.gui = "Auto"
 
+        # Use IPython if available
+        self.ipython = "yes"
```
almarklein: Why add this?

ysalmon: It is not required by the rest of the PR. I just think it is a better default. Currently, when Pyzo autodetects a config, it does not set it to use IPython, so the user has to activate it manually. But if IPython is available on the system, is it not sound to assume it is to use it ?

pyzo/core/shellStack.py
ysalmon: No. Imagine there are one or several configs but none if them was autogenerated, and we are not in mode 2 : we should not generate a new config, but ``config_to_use`` will be None.

pyzo/core/kernelbroker.py
almarklein: I prefer it not to be the default, because of backwards compat, but also because the integration has some rough edges (running a script, debugging).

pyzo/core/shellStack.py
almarklein: Right, my bad.

pyzo/core/kernelbroker.py
ysalmon: Ah, that's a little annoying because presently autodetection just will not activate IPython even in a context where this works like a charm.

Maybe the solution would be to not have hardcoded defaults for new environments in __init__ but outsource this to a config template in defaultConfig/siteConfig ? Thus a site admin could tune the template (ie. IPython but for GUI integration there might be some places where "auto" is not the best default).

almarklein: I'm not sure if I follow. Does the new autodetection in this PR behave differently regarding IPython integration? Or do you mean that at your institution you'd prefer to to make check that box by default?

almarklein:
Closing as this is now outdated.

PR #693, opened 2020-02-27
detecting if we're completing after a dot

### LINKED ISSUE #689: Do not start autocompletion from first character

ysalmon:
fixes #689 
I am not sure I did it in an entirely good way but this seems to work.


almarklein:
Thanks!

PR #694, opened 2020-03-02
Dont touch fontconfig file

### LINKED ISSUE #685: Disable FONTCONFIG_FILE env variable for kernels
We set `os.environ['FONTCONFIG_FILE']` [here](https://github.com/pyzo/pyzo/blob/master/pyzo/__init__.py#L63). Supposedly it was [added](https://github.com/pyzo/pyzo/commit/e1266e0e9f5f91fa0a0c2331f2da9399933b57aa#diff-94822694969a9a97f828619d86d1e5ad) to "make frozen Pyzo work for Linux again" in sept 2016. That was before we used PyInstaller.

We should check whether this is still needed at all. If not, let's just remove it. Otherwise, we should strip that variable from the env that is provided to the kernels (in `kernelbroker.py`).

cc @Tontyna

Tontyna:
Should close #685


almarklein:
Thanks!

PR #705, opened 2020-06-10
fix #704: better handling of carriage returns in shell

### LINKED ISSUE #704: Carriage return not handled properly in shell window
With the following, I'd expect to see "text1" on the first line, followed after a while by "text2" on the second line, which gets overwritten on the second line by "text3", which then gets overwritten on the second line by "text4" (this is the behaviour in normal Linux shell). Instead, in the Pyzo shell window, "text1" appears on the first line, then disappears, "text2" and "text3" never appear, and at the end "text4" appears on the first line.

```
from time import sleep

print('text1')
sleep(0.5)
print('text2', end='\r')
sleep(0.5)
print('text3', end='\r')
sleep(0.5)
print('text4')
sleep(0.5)
```

This might seem like a niche issue, but it prevents e.g. shell-based progressbars like [fastprogress](https://github.com/fastai/fastprogress) from working within the Pyzo shell.

I'm looking into how best to fix this (maybe the fix is easy), and will file PR if/when I fix it. Meanwhile, any pointers/input are very welcome!

Spaak:


pyzo/core/shell.py
```diff
@@ -728,12 +744,17 @@ def _handleCarriageReturn(self, text):
...
+            and not text[1:].startswith("\n")
+            and not self._lastline_had_lf
         ):
-            cursor.movePosition(cursor.PreviousBlock, cursor.KeepAnchor, 1)
+            cursor.movePosition(cursor.StartOfLine, cursor.KeepAnchor, 1)
             cursor.removeSelectedText()
         # Is this new line ending in CR?
         self._lastline_had_cr = text.endswith("\r")
+        # Is this new line ending in LF?
+        self._lastline_had_lf = text.endswith("\n")
```
almarklein, line 756: Could you please init this attribute where we also init `self._lastline_had_cr`?

almarklein:
Tiny comment, but otherwise this looks good, and a very useful fix indeed!

almarklein:
Thanks!

Spaak:
> Thanks!

You're very welcome :)

PR #712, opened 2020-09-08
Correct Pyzo with PySide2 segmentation fault

Diti24:
When using Pyzo with PySide2 when closing the IDE we get a Segmentation fault
This patch is a work around for the issue

### DIFF
--- a/pyzo/core/main.py
+++ b/pyzo/core/main.py
@@ -238,7 +238,7 @@ class MainWindow(QtWidgets.QMainWindow):
         # Set
         self.setWindowTitle(title)
 
-     def saveWindowState(self):
+    def saveWindowState(self):
         """ Save:
             * which tools are loaded
             * geometry of the top level windows
--- a/pyzo/core/main.py
+++ b/pyzo/core/main.py
@@ -238,7 +238,7 @@ class MainWindow(QtWidgets.QMainWindow):
         # Set
         self.setWindowTitle(title)
 
-    def saveWindowState(self):
+     def saveWindowState(self):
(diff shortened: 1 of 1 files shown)


pyzo/core/main.py
```diff
@@ -252,19 +252,21 @@ def saveWindowState(self):
...
         geometry = self.saveGeometry()
         try:
-            geometry = bytes(geometry)  # PyQt4
+            #geometry = bytes(geometry)  # PyQt4
+            pass
         except Exception:
             geometry = bytes().join(geometry)  # PySide
-        geometry = base64.encodebytes(geometry).decode("ascii")
+        geometry = base64.encodebytes(geometry.data()).decode("ascii")
         pyzo.config.state.windowGeometry = geometry
```
almarklein, line 260: So this makes the try-except empty, because there is a single approach for both PyQt5 and PySide2, right? Lets remove the try-except too then, please :)

almarklein:
Thanks! Just a minor comment.

pyzo/core/main.py
Diti24: Hi, so unfortunately I do not have sufficient knowledge to cleanly solve this issue. I thing someone with deeper knowledget of PyQt5/PySide2 should look at this.
With this "temporar" solution at least it does not Segmentation Fault.

almarklein: Ok, no problem. In that case I'll merge and do a bit of cleanup directly afterwards.

almarklein:
Cleaning up in #716

Issue #713, opened 2020-09-13
you need to replace deprecated ".isAlive" per ".is_alive" in two places for python-3.9rc1 compatiblity


stonebig:
I fail to start Pyzo, fisrt time I see this. any idea ?

> WPy64-3902\scripts>pyzo
> Traceback (most recent call last):
>   File "WPy64-3902\python-3.9.0rc1.amd64\lib\runpy.py", line 197, in _run_module_as_main
>     return _run_code(code, main_globals, None,
>   File "WPy64-3902\python-3.9.0rc1.amd64\lib\runpy.py", line 87, in _run_code
>     exec(code, run_globals)
>   File "WPy64-3902\python-3.9.0rc1.amd64\Scripts\pyzo.exe\__main__.py", line 4, in <module>
>   File "WPy64-3902\python-3.9.0rc1.amd64\lib\site-packages\pyzo\__init__.py", line 81, in <module>
>     if commandline.is_our_server_running():
>   File "WPy64-3902\python-3.9.0rc1.amd64\lib\site-packages\pyzo\core\commandline.py", line 150, in is_our_server_running
>     return server and server.isAlive()
> AttributeError: 'Server' object has no attribute 'isAlive'
> 
> WPy64-3902\scripts>

stonebig:
package list:

````
Package                       Version
----------------------------- ----------------
adodbapi                      2.6.1.3
affine                        2.3.0
aiofiles                      0.5.0
aiohttp                       3.6.2
aiosqlite                     0.15.0
alabaster                     0.7.12
altair                        4.1.0
altair-data-server            0.4.1
altair-transform              0.2.0
altair-widgets                0.2.2
altgraph                      0.17
amply                         0.1.2
aniso8601                     8.0.0
ansiwrap                      0.8.4
appdirs                       1.4.4
argon2-cffi                   20.2.0.dev0
asciitree                     0.3.3
asgi-csrf                     0.6.1
asgiref                       3.2.10
asteval                       0.9.19
astroid                       2.4.2
astroML                       0.4.1
astropy                       4.0.1
async-generator               1.10
async-timeout                 3.0.1
atomicwrites                  1.4.0
attrs                         20.2.0
autopep8                      1.5.3
Babel                         2.8.0
backcall                      0.2.0
backports-abc                 0.5
baresql                       0.7.6
beautifulsoup4                4.9.1
black                         19.10b0
bleach                        3.1.5
blinker                       1.4
blosc                         1.9.1
[...]

PR #714, opened 2020-09-13
replace deprecated isAlive() per is_alive()

stonebig:
deprecated isAlive() has been removed in Python-3.9

solves https://github.com/pyzo/pyzo/issues/713


almarklein:
Thanks for digging in and providing a fix!

Issue #528, opened 2018-02-27
Add "toggle comment"


taiya:
Most IDEs (with the exception of MATLAB, which does the same as Pyzo) allow commenting and uncommenting with a single shortcut (for example "CMD+/" in Textmate). In Pyzo you are forced to select two different shortcuts. It would be great if you could add a **Edit/Toggle comment** to resolve the problem?

almarklein:
That actually makes a lot of sense :)

almarklein:
Care for a PR?

mullenkamp:
I would love to see this too! I'm switching over from Spyder.

l33tlinuxh4x0r:
ctrl + 3
or 
ctrl + # 
would be preferred for me... I might look into this.

l33tlinuxh4x0r:
I just made a pull request that fixes this bug.

almarklein:
Closed by #718, thanks to @l33tlinuxh4x0r

l33tlinuxh4x0r:
I have found an issue with my first revision of code. It literally toggles comments. For example:
```#line

#with

#comment

#would

#swap
```
would become
```line
#
with
#
comment
#
would
#
swap
```

I just cleaned up the code and implemented a fix I am going to lint it and then make another pull request.

PR #718, opened 2020-09-26
Adds Toggle Comment for code editor.

l33tlinuxh4x0r:
Adds Toggle Comment for code editor and fixes bug #528


almarklein:
Looks great!

I was about to ask why "Ctrl+3", when I looked at my keyboard to realize it :)  So for future reference: instead of `Shift+3` to add "#", you can now `Ctrl+3` to toggle it.

almarklein:
Thanks for implementing the nice feature!

PR #739, opened 2020-12-17
Fixing lupdate problems and improving russian translation

f0ma:
This PR fix problem with skipping translations strings with last comma symbol by lupdate, added some missing translations and significantly improve russian translation (translated all but pyzoInteractiveHelp).


almarklein:
Thanks for this work!

There is one problem with regard to the trailing comma: that comma is being added by `black` the tool we use to auto-format the code. What is the problem with that comma exactly? Perhaps we can fix it in another way ...

f0ma:
Problem is more complicated, then I think before. `pylupdate` call  minimalistic parser for .py files processing [(this code?)](https://github.com/baoboa/pyqt5/blob/master/pylupdate/fetchtr.cpp). This parser reject `translate` calls with comma before closed bracket as invalid. `black` have a [strict policy](https://black.readthedocs.io/en/stable/the_black_code_style.html#trailing-commas) about comma and it ok in terms of Python syntax. I tried some syntax variations but without effect (`pylupdate` not working or `black` check failing). Possible solution — using custom Python AST based pylupdate (some source can be found (there)[https://github.com/pathmann/pyTSon/blob/master/tools/pylupdate.py]).

f0ma:
I found that in [PyQt6](https://riverbankcomputing.com/software/pyqt/download) pylupdate completely reworked. Now it use AST based parser (and have no problems with tailing comma), but it does not support `translate` as a function call (only ugly `QT_TRANSLATE_NOOP`), only as a method call e.g. `pyzo.translate` (but this can be relatively easy fixed). Also now pylupdate not using `.pro` file for source and `.ts` discovery. It require explicit translations list and source file list or directory for recursive walking. It is acceptable to update all translations file with patched PyQt6 lupdate? Or we should find another solutions?

almarklein:
Thanks for looking into this. This is a problem that needs to be fixed. 

I'll probably have much more time to work on Pyzo in the coming weeks, and this will definitely be something I'll look into.

almarklein:
This PR is outdated. I'm also not sure anymore what the problem is. Translations seem to still work with PySide6.4 at least.

PR #758, opened 2021-09-07
new black

almarklein:


azure-pipelines.yml
```diff
@@ -67,7 +67,7 @@ jobs:
           apt-get update
           apt-get install -y python3 python3-dev python3-pip python3-pyqt5 &&\
           python3 -m pip install -U pip &&\
-          python3 -m pip install -U pyinstaller &&\
+          python3 -m pip install -U https://github.com/pyinstaller/pyinstaller/archive/develop.zip &&\
```
almarklein, line 70: In this PR I kept getting an error that PyQt5 does not have a `__version__` attribute, with the Docker build that uses `apt install python3-pyqt5`.

This bug seems to have been introduced relatively recently, and has also been fixed in the `develop` branch:

https://github.com/pyinstaller/pyinstaller/blob/0ffb67eddbdb8aa3a49101e5eb1dcc3ecd484f10/PyInstaller/utils/hooks/qt.py#L44-L52

At the moment this fix is not yet released (current latest release is 4.5.1 from 07-08-2021).

almarklein: If we merge this, I should make an issue to undo this particular change.

PR #778, opened 2021-12-20
Refactor freezing process

almarklein:
This is a major refactoring of how the freezing process works. 
* [x] CI now has a test that runs Pyzo for 5 seconds.
* [x] CD produces binaries and uploads to a release when a tag is pushed.  
* [x] Refactored Pyzo's `__init__`, moving the majority to `_start.py`.
* [x] Only importing `pyzo` does not yet import qt or any gui stuff yet.
* [x] Freeze includes are declared explicitly.
* [x] The freezing includes (almost) the whole stdlib. This is to make it possible to update Pyzo using the same frozen base (in the future).
* [x] Now running on Python 10 and PySide6 for Win65 and MacOS. Using py39 and PyQt5 for Win32 and Linux.


almarklein:
Result for testing here: https://github.com/almarklein/pyzo/releases/tag/vt201 Please let me know whether it works for you!

Issue #796, opened 2022-03-10
'IPython.utils.io' has no attribute 'IOStream'


UniqASL:
Hi,

The last version of IPython 8.1.1 is not working anymore in Pyzo (4.12.0, Windows 10), I'm getting the following error:

> IPython could not be loaded: module 'IPython.utils.io' has no attribute 'IOStream'

Downgrading to IPython 7.32.0 solves the problem.

I [have been asking IPython](https://github.com/ipython/ipython/issues/13576) and the problem seem to come from Pyzo and not from their side.

Any idea how this problem can be solved?

almarklein:
It looks like their API has changed and we need to adjust accordingly.

almarklein:
I looked into this, but it looks like the factored that class away, and now I don't know how to fix it. It also looks like IPython integration is broken on MacOS for another reason. 

If anyone wants to look into it, this is the relevant code for this issue:
https://github.com/pyzo/pyzo/blob/ba3d7ebb3953f9cb83c430f467f54ef0f42e6369/pyzo/pyzokernel/interpreter.py#L591-L599

It could be that IPython has changed in a way that this fix-for-Windows is no longer needed. You can try editing this file on your local machine (see "help > about pyzo" for the location of your source directory).

UniqASL:
Hi,

thanks for your help. I slightly changed the code and created a [pull request](https://github.com/pyzo/pyzo/pull/803).

So far it is working without problem, IPython 8.1.1 is loading.

Cheers

PR #803, opened 2022-03-17
Bug IOStream from IPython 8.x.x

UniqASL:
On Windows, IPython is not loaded from version 8.x.x and the console throws the error:

> 'IPython.utils.io'` has no attribute 'IOStream'

See [issue #796](https://github.com/pyzo/pyzo/issues/796).

So far the fix is working without problem.

### DIFF
--- a/pyzo/pyzokernel/interpreter.py
+++ b/pyzo/pyzokernel/interpreter.py
@@ -592,11 +592,13 @@ class PyzoInterpreter:
             # IPython wraps std streams just like we do below, but
             # pyreadline adds *another* wrapper, which is where it
             # goes wrong. Here we set it back to bypass pyreadline.
-            from IPython.utils import io
+            
+            if IPython.version_info < (8,): #corrects a problem with IOStream from IPython 8.x.x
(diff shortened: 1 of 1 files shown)


almarklein:
Thanks!

When I looked at this, I tried to find a way to wrap the pipes for newer IPython's. But indeed maybe it's not necessary for newer versions. Let's try and see :)

pyzo/pyzokernel/interpreter.py
```diff
@@ -592,11 +592,13 @@ def _load_ipyhon(self):
             # IPython wraps std streams just like we do below, but
             # pyreadline adds *another* wrapper, which is where it
             # goes wrong. Here we set it back to bypass pyreadline.
-            from IPython.utils import io
+            
+            if IPython.version_info < (8,): #corrects a problem with IOStream from IPython 8.x.x
```
almarklein, line 596:
```suggestion
            if IPython.version_info < (8,):  # corrects a problem with IOStream from IPython 8.x.x
```

PR #807, opened 2022-07-07
Compatibility Python 3.10

YoannUniKS:
notifyAll() deprecated and replaced with notify_all() in Python 3.10

### DIFF
--- a/pyzo/yoton/channels/channels_base.py
+++ b/pyzo/yoton/channels/channels_base.py
@@ -257,7 +257,7 @@ class BaseChannel(object):
         if not value:
             self._send_condition.acquire()
             try:
-                if sys.version_info[0] < 3.10:
+                if sys.version_info < (2, 6):
                     self._send_condition.notifyAll()
                 else:
                     self._send_condition.notify_all()
--- a/pyzo/yoton/channels/channels_base.py
+++ b/pyzo/yoton/channels/channels_base.py
@@ -9,7 +9,7 @@
 Defines the base channel class and the MessageType class.
 
 """
-
+import sys
 import time
 import threading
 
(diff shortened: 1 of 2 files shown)


pyzo/yoton/channels/channels_base.py
```diff
@@ -257,7 +257,10 @@ def _set_send_lock(self, value):
         if not value:
             self._send_condition.acquire()
             try:
-                self._send_condition.notifyAll()
+                if sys.version_info[0] < 3.10:
```
almarklein, line 260:
```suggestion
                if sys.version_info < (2, 6):
```

pyzo/yoton/misc.py
```diff
@@ -569,7 +569,10 @@ def pop(self, block=True):
 
             # Notify if this pop would reduce the length below the threshold
             if len(q) <= self._tinylen:
-                condition.notifyAll()  # wait() procedes
+                if sys.version_info[0] < 3.10:
```
almarklein, line 572:
```suggestion
                if sys.version_info < (2, 6):
```

almarklein (commented):
Thanks!

Note that version info is a tuple, so you should compare with a tuple too. Also `notify_all` is available since Python 2.6. Updated the code accordingly.

pyzo/yoton/misc.py
```diff
@@ -569,7 +569,10 @@ def pop(self, block=True):
 
             # Notify if this pop would reduce the length below the threshold
             if len(q) <= self._tinylen:
-                condition.notifyAll()  # wait() procedes
+                if sys.version_info < (2, 6):
+                    condition.notifyAll()  # wait() procedes
+                else:
+                    condition.notify_all()
```
almarklein, line 575:
```suggestion
                    condition.notify_all()
```

Issue #812, opened 2022-08-23
Source structure's background color and the font's front-color are unchangeable


Ahmed-E-86:
Source structure's background color and font front-color are unreadable, as you can see in this picture

![Screenshot_20220823_143751](https://user-images.githubusercontent.com/29783425/186162423-13bc5a99-dd8c-4cc0-9ee7-dccddf56117c.png)

Pyzo version: 4.12.3 (source)
Platform: Linux
Linux distro: Kubuntu 22.04 / KDE Plasma 5.24.6
Python version: 3.10.4
Qt version: 5.15.3
PyQt5 version: 5.15.6

I tried all the themes of the application, and all the themes of the operating system, to no avail. The background color and the font's front-color are unchangeable.  Could you please help me with this issue?

PR #813, opened 2022-08-23
Update pyzoSourceStructure.py

Ahmed-E-86:
These colors are better than the other ones.

### DIFF
--- a/pyzo/tools/pyzoSourceStructure.py
+++ b/pyzo/tools/pyzoSourceStructure.py
@@ -253,19 +253,26 @@ class PyzoSourceStructure(QtWidgets.QWidget):
         ln += 1  # is ln as in line number area
 
         # Define colours
+        # colours = {
+        #     "cell": "#b58900",
+        #     "class": "#cb4b16",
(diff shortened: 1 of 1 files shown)


Ahmed-E-86:
This PR solves this issue -> https://github.com/pyzo/pyzo/issues/812

almarklein:
I don't see how this fixed #812, does it not still have a light background? Could you show a screenshot?

Ahmed-E-86:
It is the same background, but with brighter colors.

Actually they are the same colors that you commented before, I just uncommented them, and they look better with dark background, and I think that they will be fine with light background as well.

Screenshot_20220907_152123

almarklein:
Now I see! The text is barely visible, and the lighter background of the top item is because it is selected.

> I think that they will be fine with light background as well.

Well, the colors currently match the default syntax highlighting, so this change would introduce a discrepancy. The best solution would be if the source structure simply took the colors from the syntax theme. This is a bit more work though ...

almarklein:
Supersceded by #828

Issue #816, opened 2022-10-21
runaway memory leak


don6019:
Version info
Pyzo version: 4.12.3 (source)
Platform: linux
Python version: 3.9.12
Qt version: 5.15.2
PyQt5 version: 5.15.7
==================================================
I have a problem with Pyzo when using a particular python library - tinydb

In the pyzo shell - 

`>>> pip install tinydb
Collecting tinydb
  Using cached tinydb-4.7.0-py3-none-any.whl (24 kB)
Installing collected packages: tinydb
Successfully installed tinydb-4.7.0

`>>> from tinydb import TinyDB, Query

`>>> Q = Query()

`>>> `

This causes the python process to rapidly consume growing amounts of memory until the shell is restarted (before the system runs out of memory and freezes).
Same thing happens if I run the above in a script in the Pyzo editor.

If I start the python interpreter in a separate terminal window this does not happen, also does not happen in other IDE such as IDLE or Eclipse - only in Pyzo - strange ??

I have found the same problem using Pyzo on Windows (10) and Linux (ZorinOS 16.1)

Curious that it is only this particular library which causes a problem with Pyzo but nowhere else ??

I really love using Pyzo - it is far and away the best Python IDE I have found and use it almost everyday. Can anyone else duplicate this issue and or come up with a solution or work around.

Thanks for reading.

Don

almarklein:
> I really love using Pyzo - it is far and away the best Python IDE I have found and use it almost everyday.

❤️ 

> This causes the python process to rapidly consume growing amounts of memory until the shell is restarted (before the system runs out of memory and freezes).

That is strange indeed. I just tested on MacOS, and cannot reproduce. Will test on Windows later.

Do you have a GUI integration on? If so, could you try turning that to `None` in the shell config?

don6019:
Thanks for the reply.
Yes I did try with GUI set to None (and also some of the others). That didn't appear to make any difference unfortunately.

don6019:
I tried the modified introspection.py as proposed by adddvent and this does indeed appear to have solved my problem with TinyDB and Pyzo.

Thankyou very much.

almarklein:
Closed by #818

almarklein:
Thanks for reporting this and testing the proposed solution!

PR #818, opened 2022-10-21
fix introspection infinite iteration for special cases

adddvent:
As described in #816, some special cases will cause the Pyzo introspection to iterate infinitely for special class objects, and this can result in steadily rising memory usage.

The example in the issue can be broken down to the code
```
class Query:
    def __init__(self):
        print('__init__')

    def __getattr__(self, item):
        print('__getattr__', item)
        return self

    def __getitem__(self, item):
        print('__getitem__', item)
        # TODO: allocate memory
        return

Q = Query()
```
When running this code in Pyzo, pyzokernel/introspection.py will see a new global variable "Q" and run "dir2". Then "dir2" and "storeInfo" will execute `hasattr(val, "__array__")` etc. and come to the conclusion, that the object is an array -- because the Query class will have any attribute that is looked up (see `__getattr__`). Still in "storeInfo", the line `tmp = "x".join([str(s) for s in val.shape])` is executed to make a string that contains e.g. "2x3" for a 2 rows 3 columns array, but in this case, `val.shape` will be a `Query` object. And inside the list comprehension, `val.shape[0]`, `val.shape[1]`, `val.shape[2]`, etc. will be accessed in an infinite loop because `Query.__getitem__` never gets exhausted.

The proposed change fixes that special case by adding another criterion for checking if the object is an array.

### DIFF
--- a/pyzo/pyzokernel/introspection.py
+++ b/pyzo/pyzokernel/introspection.py
@@ -254,7 +254,7 @@ class PyzoIntrospector(yoton.RepChannel):
                         hasattr(val, "__array__")
                         and hasattr(val, "dtype")
                         and hasattr(val, "shape")
-                        and not hasattr(val, "__this_object_is_definitely_not_an_array__")
+                        and not hasattr(val, "if_this_is_an_attribute_then_there_are_likely_inf_attributes")
                     ):
                         kind = "array"
                     elif isinstance(val, list):
--- a/pyzo/pyzokernel/introspection.py
+++ b/pyzo/pyzokernel/introspection.py
@@ -254,6 +254,7 @@ class PyzoIntrospector(yoton.RepChannel):
                         hasattr(val, "__array__")
                         and hasattr(val, "dtype")
                         and hasattr(val, "shape")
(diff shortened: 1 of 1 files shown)


pyzo/pyzokernel/introspection.py
```diff
@@ -254,6 +254,7 @@ def storeInfo(name, val):
                         hasattr(val, "__array__")
                         and hasattr(val, "dtype")
                         and hasattr(val, "shape")
+                        and not hasattr(val, "__this_object_is_definitely_not_an_array__")
```
almarklein, line 257: A small tweak so that a future reader understand why this case is here. I actually looked up `__this_object_is_definitely_not_an_array__` on Google because my initial thought was that this is a convention 😆 
```suggestion
                        and not hasattr(val, "if_this_is_an_attribute_then_there_are_likely_inf_attributes")
```

almarklein:
CI failure is due to an unrelated problem in PySide6.

Thanks a lot for digging into this and finding a solution! ❤️

Issue #814, opened 2022-09-03
Force a specific python executable declaratively


SCOTT-HAMILTON:
Hello, I'm using NixOS and I'm packaging pyzo. The strength of Nix is to make everything declarative, so I'm trying to declaratively configure pyzo through nix to make it use the python executable I want, thus being able to leverage the excellent Nix's python ecosystem.

Therefore I'm asking if it's possible to force pyzo to use a specific python binary. Either with:
 - Environment variable
 - CLI
 - Config (not the best)
Note: If the only way is through a config file, then this file will have to be read-only because if nix produces it, then it's a declarative output which must be immutable.

If this isn't possible yet, I'm happy to try and implement whatever above solutions you like the best and open a PR if I'm successful with it.

almarklein:
>  if it's possible to force pyzo to use a specific python binary

Do you mean for the process that runs Pyzo, or for the shell, as an override for the shell configuration?

SCOTT-HAMILTON:
The use case is to be able force the pyzo IDE to use a specific python executable without having to set it manually via the menu. That would make integration with Nix much easier.

almarklein:
At the moment this does not exist, but I would be open for an ENV var or CLI arg that sets up a default shell when there are none yet.

PR #819, opened 2022-11-01
Introduce PYZO_DEFAULT_SHELL_PYTHON_EXE env-var

SCOTT-HAMILTON:
Closes https://github.com/pyzo/pyzo/issues/814

Makes pyzo consider a user-defined python exe file without having to write it down manually in the file explorer.

This comes in handy for NixOS users because they'll be able to make pyzo automatically propose a specific python executable from the nix store, which is a better UX than having to directly deal with the hash-looking paths of the store...


almarklein:
Sorry for the late reply. This got lost in my mailbox. This looks great, thanks!

The CI issues should have been fixed, so I'll just merge this.

PR #827, opened 2022-12-20
Fix that applying a theme is broken on some systems

### LINKED ISSUE #805: Cannot change theme
When I restart Pyzo (`File > Restart Pyzo`), I get an error message when Pyzo reopens:
Capture d’écran 2022-06-09 110300
I don’t get any error message if I quit Pyzo then manually restart it. However, if I go to `Settings > Edit syntax styles… `, choose a non-default theme and click on `Apply theme`, Pyzo automatically restarts, I get the error message, and the theme is still the default one. As a consequence, I cannot change the theme.

I use Windows 10 64 bits. Here is the `About Pyzo` info:
```
Version info
Pyzo version: 4.12.3 (binary)
Platform: win32
Python version: 3.10.2
Qt version: 6.2.3
PySide6 version: 6.2.3

Pyzo directories
Pyzo source directory: C:\Program Files\pyzo\source\pyzo
Pyzo config directory: C:\Users\Myvh773\AppData\Roaming\pyzo
Pyzo userdata directory: C:\Users\Myvh773\AppData\Roaming\pyzo
```

I uninstalled Pyzo, restarted Windows, reinstalled Pyzo and restarted Windows again, but the bug is still here.

almarklein:
Closes #805

### DIFF
--- a/pyzo/core/themeEdit.py
+++ b/pyzo/core/themeEdit.py
@@ -436,6 +436,7 @@ class ThemeEditorWidget(QtWidgets.QWidget):
 
         if prev != new:
             pyzo.config.settings.theme = new
+            pyzo.saveConfig()
             # This may be better
             pyzo.main.restart()
         else:


almarklein:
Not a perfect solution, because the fact that the config is not stored on restart indicates that `on_close` is not called properly ... but it'll do for now.

PR #835, opened 2023-01-20
Allow <tmp 1> etc. as filename when double-clicking text

bdieterm:
Double-clicking on a file name in the shell text, for example in the traceback message
`File "C:\temp\test.py", line 5, in <module>`
will open the file in an editor tab and move the cursor to the specified line number.
But for new, unsaved files with pseudo filenames "<tmp 1>" etc., this does not work.

The proposed change will make it possible to jump to lines of such new, unsaved files.


almarklein:
Nice find!

PR #837, opened 2023-01-25
have proper placeholder text for environ in shell config dialog

bdieterm:
When editing or adding a shell configuration using the dialog, the text box "environ" shows the following text instead of an empty string (unless modified by the user):
```
EXAMPLE_VAR1=value1
PYZO_PROCESS_EVENTS_WHILE_DEBUGGING=1
```
But these variables never appear in os.environ -- the shell config ignores the text because it is a kind of placeholder text but implemented as normal text data. This has caused some confusion, as placeholder texts in other fields in the shell config dialog are implemented as proper placeholder texts (grayed out and vanishing when entering text).

This pull request changes the environ text field to have a proper placeholder text.
The old implementation allowed for easy inserting the useful environment variable `PYZO_PROCESS_EVENTS_WHILE_DEBUGGING=1` by just modifying the default entry. With the proper placeholder text, the user would have to type the whole line manually, because placeholder texts cannot be selected and copied to the clipboard.
To circumvent this, the default evironmental variable for a new shell config is changed to include this line.

### DIFF
--- a/pyzo/core/kernelbroker.py
+++ b/pyzo/core/kernelbroker.py
@@ -106,7 +106,7 @@ class KernelInfo(ssdf.Struct):
         self.argv = ""
 
         # Additional environment variables
-        self.environ = "PYZO_PROCESS_EVENTS_WHILE_DEBUGGING=1"
+        self.environ = ""
 
         # Load info from ssdf struct. Make sure they are all strings
         if info:
--- a/pyzo/core/kernelbroker.py
+++ b/pyzo/core/kernelbroker.py
@@ -106,7 +106,7 @@ class KernelInfo(ssdf.Struct):
         self.argv = ""
 
         # Additional environment variables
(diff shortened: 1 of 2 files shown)


pyzo/core/kernelbroker.py
```diff
@@ -106,7 +106,7 @@ def __init__(self, info=None):
         self.argv = ""
 
         # Additional environment variables
-        self.environ = ""
+        self.environ = "PYZO_PROCESS_EVENTS_WHILE_DEBUGGING=1"
```
almarklein, line 109: I don't think this is needed - previously, when the text was the same as the example, it was interpreted as an empty string. 
```suggestion
        self.environ = ""
```

almarklein (commented):
Nice! Glad to remove this awkward mechanism :)

PR #842, opened 2023-02-06
Signing of the app for macos

almarklein:
## Ad-hoc code signing

In https://github.com/pyinstaller/pyinstaller/pull/7180#issuecomment-1415721750 a tip for the `codesign` syntax was given:
```
codesign -s - --force --all-architectures --timestamp --deep --entitlements /path/to/entitlements-file.plist /path/to/bundle.app
```

This works, and MacOS stops complaining about the app being broken or modified. But it still says it's untrusted, and requires right-clicking to open the app.

## Proper code signing

For proper code signing, you need to:
* Create an Apple developer account ($100 per year)
* Create a certificate with that account
* Setup Github Actions to sign with that certificate.
* Setup Github Actions to Notarization the app (verify that it's clean).

For detailed instructions, see https://federicoterzi.com/blog/automatic-code-signing-and-notarization-for-macos-apps-using-github-actions/

## Trying to do proper code signing

So I gave this a go ... it was a tedious process with many hurdles. Every time it felt like I was close, but everytime it did not work for another reason ...

For one, I changed the `codesign` signature to include more options:
```
codesign --force -s "$MACOS_CERTIFICATE_NAME" --all-architectures --timestamp --deep --options runtime freeze/dist/pyzo.app -v
```

Also, my own certificate was marked "Certificate is not trusted", but I could fix that by checking what the issuer was, and then downloading and installing the corresponding certificate from [Apple PKI](https://www.apple.com/certificateauthority/), as explained [here](https://developer.apple.com/forums/thread/705473).

## The result

In the end, everything looks good, but I still get this:
image

So people still need to right-click, select "open", and then click "open". Maybe I'll dive deeper another time.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -124,10 +124,11 @@ jobs:
         # Here we send the notarization request to the Apple's Notarization service, waiting for the result.
         echo "Notarize app"
         xcrun notarytool submit "notarization.zip" --keychain-profile "notarytool-profile" --wait
+        xcrun notarytool log c44df8eb-3087-4a3c-a5f4-9b72caceb083 --keychain-profile "notarytool-profile"
         # Finally, we need to "attach the staple" to our executable, which will allow our app to be
         # validated by macOS even when an internet connection is not available.
         echo "Attach staple"
-        xcrun stapler staple "freeze/dist/pyzo.app"
+        #xcrun stapler staple "freeze/dist/pyzo.app"
     - name: Package
       run: python freeze/pyzo_package.py
     - name: Test frozen
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -112,7 +112,7 @@ jobs:
       run: |
           python freeze/pyzo_freeze.py
           # cant get it to work //// codesign -s $MACOS_CERT_ID --force --all-architectures --timestamp --deep --options=runtime freeze/dist/pyzo.app -v
-          codesign -s --force --all-architectures --timestamp --deep freeze/dist/pyzo.app
+          codesign -s - --force --all-architectures --timestamp --deep freeze/dist/pyzo.app
     - name: Package
       run: python freeze/pyzo_package.py
     - name: Test frozen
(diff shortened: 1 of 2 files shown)


almarklein:
Heeey, when I just downloaded the dmg and opened Pyzo from there, I got
image

So it looks like it worked??

almarklein:
Aha, it looks like when the app is packged in a dmg, it works, but not when packaged in a zip file. Perhaps some of the signing info is lost when it's zipped?

phw:
On macOS 10.14.5 and later it is also required that the software gets "notarized" by Apple. Basically the signed app gets sent to Apple for an automated malware scan. See also https://developer.apple.com/developer-id/ for some info. I think the missing notarization is what's causing the malware warning dialog.

This process is a bit more difficult to automate, but have a look how we handle that for MusicBrainz Picard at https://github.com/metabrainz/picard/blob/master/scripts/package/macos-notarize-app.sh

almarklein:
@phw The notarization is done in the action :)  - but thanks for the link, good to see alternative solutions.

To be clear, the application distributed in the .dmg now works correctly, without warnings. The weird thing is that the one distributed in a zip-file still gives the warning, even though it contains the exact same pyzo.app. My guess is file attributes get lost during the zipping.

bwoodsend:
What about a tar file? tar preserves file permissions unlike zip. Also does your application contain symlinks? zip removes them by default too. Try zip's `-l` flag.

almarklein:
> To avoid the file signatures in attributes all files that cannot be signed as part of the file itself should be put into the Resources folder instead of Contents.

That's what I do now. (if you mean `Contents/Resources` vs `Conrtents/MacOS`).

> What about a tar file? tar preserves file permissions unlike zip. Also does your application contain symlinks? zip removes them by default too. Try zip's -l flag.

I tried preserving symlinks with .zip, as well as using a .tar.gz, and in both cases I still get the warning. I'm not too worried, since by far most users will use the .dmg. Though if you have more ideas, I'll give it a shot.

jellespijker:
Thx @almarklein Thought I posted some of our methods as well, 

We (UltiMaker Cura) had to change stuff a bit more extensive, doing some duck typing, to ensure that our files are stored to the correct locations: https://github.com/Ultimaker/Cura/blob/5970524b3a6dca81c83156d9c48c65facb9481ba/UltiMaker-Cura.spec.jinja#L72 This jinja template is filled out with our dependency manager Conan here: https://github.com/Ultimaker/Cura/blob/5970524b3a6dca81c83156d9c48c65facb9481ba/conanfile.py#L183

Code signing and notarizing is done in this workflow: https://github.com/Ultimaker/Cura/blob/main/.github/workflows/cura-installer.yml

Which basically uses the scripts in this folder: https://github.com/Ultimaker/Cura/tree/main/packaging/MacOS

Issue #843, opened 2023-02-25
Trying to launch edit key mapping option crashes pyzo (pyqt5)


Pavel-Sayekat:
Trying to launch edit key mapping option crashes pyzo.

OS - Parabola Arch Linux
arch: x86_64

    Pyzo - Python to the people!

    Version info
    Pyzo version: 4.12.4 (source)
    Platform: linux
    Python version: 3.10.9
    Qt version: 5.15.8
    PyQt5 version: 5.15.7

crash log:

    $ pyzo
    Our command server is *not* running
    Loaded theme 'Default'
    Loaded theme 'Dark'
    Loaded theme 'Solarized_Dark'
    Loaded theme 'Scintilla'
    Loaded theme 'Solarized_Light'
    Fatal Python error: Segmentation fault

    Thread 0x00007f536effd6c0 (most recent call first):
      File "/usr/lib/python3.10/site-packages/pyzo/yoton/core.py", line 84 in can_recv
      File "/usr/lib/python3.10/site-packages/pyzo/yoton/connection_tcp.py", line 620 in _run
      File "/usr/lib/python3.10/site-packages/pyzo/yoton/connection_tcp.py", line 533 in run2
      File "/usr/lib/python3.10/site-packages/pyzo/yoton/connection_tcp.py", line 510 in run
      File "/usr/lib/python3.10/threading.py", line 1016 in _bootstrap_inner
      File "/usr/lib/python3.10/threading.py", line 973 in _bootstrap

    Thread 0x00007f536f7fe6c0 (most recent call first):
      File "/usr/lib/python3.10/threading.py", line 324 in wait
      File "/usr/lib/python3.10/site-packages/pyzo/yoton/misc.py", line 564 in pop
      File "/usr/lib/python3.10/site-packages/pyzo/yoton/connection_tcp.py", line 575 in _run
      File "/usr/lib/python3.10/site-packages/pyzo/yoton/connection_tcp.py", line 533 in run2
      File "/usr/lib/python3.10/site-packages/pyzo/yoton/connection_tcp.py", line 510 in run
      File "/usr/lib/python3.10/threading.py", line 1016 in _bootstrap_inner
      File "/usr/lib/python3.10/threading.py", line 973 in _bootstrap

    Thread 0x00007f536ffff6c0 (most recent call first):
      File "/usr/lib/python3.10/site-packages/pyzo/core/kernelbroker.py", line 759 in run
[...]

PR #847, opened 2023-03-29
avoid crash in key mapping settings

bdieterm:
Warning! This is just a fix to avoid the crash. I have not tested it for secondary effects.

As described in #843, calling the key mapping settings dialog will crash pyzo when using a linux os, unless the font list was loaded before.

The segmentation fault happens in method `hasChildren` in class `KeyMapModel`, probably because of an invalid memory access in line `result = isinstance(index.internalPointer(), QtWidgets.QMenu)`.

It does not seem possible to avoid a crash when already inside `hasChildren`.
Changes have to be done to method `index` of the same class.
Inserting the lines
```
if isinstance(childMenu, FontMenu):
    childAction = QtWidgets.QAction()
```
before `return self.createIndex(row, column, childAction)` still leads to the same crash.
But, when giving the dummy action object a safe parent that keeps its memory ownership
```
if isinstance(childMenu, FontMenu):
    FontMenu.asdfasdf = QtWidgets.QAction()
    childAction = FontMenu.asdfasdf
```
the crash does not occur anymore and the menu entry is displayed like a separator in the key settings dialog.
So it seems that a `QAction` object got garbage collected and the later access in `hasChildren` caused the segmentation fault.

The proposed fix could be extended to also hide other menus by inserting line `_hide_in_key_map_dialog = True` in the class definition.

### DIFF
--- a/pyzo/core/menu.py
+++ b/pyzo/core/menu.py
@@ -186,6 +186,8 @@ class Menu(QtWidgets.QMenu):
 
     """
 
+    _dummyActionForHiddenEntryInKeyMapDialog = QtWidgets.QAction()
+
     def __init__(self, parent=None, name=None):
         QtWidgets.QMenu.__init__(self, parent)
 
@@ -851,6 +853,8 @@ class ZoomMenu(Menu):
 
 
 class FontMenu(Menu):
+    _hide_in_key_map_dialog = True
+
     def __init__(self, parent=None, name="Font", *args, **kwds):
         Menu.__init__(self, parent, name, *args, **kwds)
         self.aboutToShow.connect(self._updateFonts)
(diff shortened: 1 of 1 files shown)


pyzo/core/menu.py
```diff
@@ -2652,6 +2659,9 @@ def index(self, row, column, parent):
         childMenu = childAction.parent()
         if childMenu is not parentMenu and childMenu.actions():
             childAction = childMenu
+        if (isinstance(childMenu, Menu) and hasattr(childMenu, '_hide_in_key_map_dialog')
+                and childMenu._hide_in_key_map_dialog):
```
almarklein, line 2663:
```suggestion
        if (isinstance(childMenu, Menu) and getattr(childMenu, '_hide_in_key_map_dialog', False):
```

pyzo/core/menu.py
```diff
@@ -2652,6 +2659,8 @@ def index(self, row, column, parent):
         childMenu = childAction.parent()
         if childMenu is not parentMenu and childMenu.actions():
             childAction = childMenu
+        if (isinstance(childMenu, Menu) and getattr(childMenu, '_hide_in_key_map_dialog', False):
```
almarklein, line 2662:
```suggestion
        if isinstance(childMenu, Menu) and getattr(childMenu, '_hide_in_key_map_dialog', False):
```

almarklein:
Thanks! Looks good to me - let's just try this!

PR #850, opened 2023-04-29
enable autocompletion for string literals

bdieterm:
This PR enables autocompletion when typing string literals (and similar objects), for example:
typing `' '.j` wll suggest `join`
typing `b'test'.d` wll suggest `decode`

And this PR also includes a minor code clean-up:
`isinstance(token, (Tokens.IdentifierToken, Tokens.KeywordToken))`
is the same as
`isinstance(token, Tokens.IdentifierToken)`
because `Tokens.KeywordToken` is a subclass of `Tokens.IdentifierToken`


almarklein:
Nice, thanks!

CI: Looks like PySide is having a bad day again. I'll look into that later.

PR #851, opened 2023-05-06
fix crash of kernel when importing PySide6

bdieterm:
### Description of the problem

python environment running in the shell:
os: Linux
exe: Python 3.11.1 
gui: PySide6 6.5.0

When creating or restarting the shell, it crashes:
```
Fatal Python error: Segmentation fault
Python 3.11.1 (main, Jan 16 2023, 22:41:20) on linux (64 bits).
...
Extension modules: xxsubtype, shiboken6.Shiboken (total: 2)
Segmentation fault (core dumped)

The kernel process exited. (139)
```

### Cause of the problem
Importing `PySide6` will also import module `shiboken6`.
And `shiboken6` replaces python's `__import__` with a custom C function `__feature_import__`.
For every module that is imported as a text file (*.py), shiboken calls its own function `_mod_uses_pyside` and in that function `source = inspect.getsource(module)` is executed.
And python's `inspect.getsource` will call python's `getlines`.
But `getlines` is monkey-patched in `class ExecutedSourceCollection` in `pyzo/pyzokernel/interpreter.py`
and inside the patched `getlines` there is the command `import linecache`.
That import command will again call the `__feature_import__` from shiboken6 and the infinite loop is closed which causes the recursion limit resp. the segmentation fault.

`shiboken6` is a binary module but the source code of the critical part with the import override can be seen at
[https://code.qt.io/cgit/pyside/pyside-setup.git/tree/sources/shiboken6/shibokenmodule/files.dir/shibokensupport/feature.py](https://code.qt.io/cgit/pyside/pyside-setup.git/tree/sources/shiboken6/shibokenmodule/files.dir/shibokensupport/feature.py)

### Solution
In `pyzo/pyzokernel/interpreter.py`, the command `import linecache` is removed from the patched `getlines` and added on the top of the file, next to the other imports. I am not aware of any negative side effects.

### DIFF
--- a/pyzo/pyzokernel/interpreter.py
+++ b/pyzo/pyzokernel/interpreter.py
@@ -36,6 +36,7 @@ import keyword
 import inspect  # noqa - Must be in this namespace
 import bdb
 from distutils.version import LooseVersion as LV
+import linecache
 
 import yoton
 from pyzokernel import guiintegration, printDirect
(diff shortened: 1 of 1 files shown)


almarklein:
Nice digging! Thanks a lot! 👍 🚀

Issue #852, opened 2023-05-06
missing package for CI workflow for Qt 6.5 in linux


bdieterm:
Qt 6.5 has an additional dependency in linux:

In the CI workflow, package "libxcb-cursor0" has to be added, like this:
```
    - name: Prepare run
      if: matrix.os == 'ubuntu-latest'
      run: |
        sudo apt install -y libdbus-1-3 libxkbcommon-x11-0 libxcb-icccm4 \
            libxcb-image0 libxcb-keysyms1 libxcb-randr0 libxcb-render-util0 \
            libxcb-xinerama0 libxcb-xfixes0 libxcb-shape0 libxcb-cursor0
```

PR #853, opened 2023-05-06
added missing linux packet for PySide 6.5

bdieterm:
changes for issue #852


almarklein:
Thanks for the fix! 🚀

PR #858, opened 2023-05-10
fix pdf export

bdieterm:
PDF export works now across PyQt and PySide with Qt5 and Qt6. And page orientation can be set to landscape again.

worth mentioning:
Opening the PDF export dialog when running Pyzo with PySide2 Qt bindings in Linux caused a segmentation fault. Changing `PdfExport().exec()` to `PdfExport().exec_()` solved that. All other dialogs (subclasses of `QtWidgets.QDialog`) already used the unproblematic `.exec_()` call.


almarklein:
Nice!

Issue #859, opened 2023-05-12
Application aborted while Runnig Pyzo


cdavailable:
Hi everyone! I could use your help please.

Before when a file was open on the Pyzo window, and I double clicked on another .py file, it opened directly on the Pyzo editor. I installed the new version of Pyzo and when I do the same operation, I get a cmd window that opens and closes several times before displaying this message "Application aborted while Running Pyzo. SystemExit:" And then, the file opens in the editor.

Do you have any idea how to fix this problem. Thank you very much for your help.

bdieterm:
Hi @cdavailable!
PR #860 will fix this problem.

almarklein:
Closed by #860

PR #860, opened 2023-05-14
prevent error message box when running a second Pyzo instance to open a file

bdieterm:
As described in #859, starting a second instance of Pyzo to open a file will show a cmd console window and an error message box in MS Windows.
The console window and the error message box seem to be opened because an exception is raised after successfully opening the file. This exception is of type `SystemExit` and caused by the command
https://github.com/pyzo/pyzo/blob/ebb7df8922af8e15eac102e8fa4a35bfa143869a/pyzo/_start.py#L77
which is just called to gracefully close the second Pyzo instance.

### DIFF
--- a/freeze/boot.py
+++ b/freeze/boot.py
@@ -68,8 +68,10 @@ class BootAction:
 
     def __exit__(self, cls, err, tb):
         if err:
-            error_handler(cls, err, tb, self._action)
-            sys.exit(1)
+            if not isinstance(err, SystemExit) or err.code:
+                error_handler(cls, err, tb, self._action)
+                if not isinstance(err, SystemExit):
+                    sys.exit(1)
 
 
 # %% Boot
--- a/pyzo/__init__.py
+++ b/pyzo/__init__.py
@@ -52,13 +52,6 @@ if sys.version_info < (3, 6):
 
 def start():
     """Start Pyzo."""
-    try:
-        from ._start import start
-    except SystemExit as e:
-        if e.code == 0 or e.code is None:
-            # prevent SystemExit exception from showing an error message box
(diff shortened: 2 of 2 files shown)


almarklein:
Thanks for tracking this down!

I think the cleaner way to solve this would be the `__exit__` in this code:
https://github.com/pyzo/pyzo/blob/ebb7df8922af8e15eac102e8fa4a35bfa143869a/freeze/boot.py#L58-L72

bdieterm:
Thanks for the hint.
I have updated the code changes.

freeze/boot.py
```diff
@@ -68,8 +68,10 @@ def __enter__(self):
 
     def __exit__(self, cls, err, tb):
         if err:
-            error_handler(cls, err, tb, self._action)
-            sys.exit(1)
+            if not isinstance(err, SystemExit) or err.code:
+                error_handler(cls, err, tb, self._action)
+                if not isinstance(err, SystemExit):
+                    sys.exit(1)
```
almarklein, line 74: Thanks!  I don't think all exceptions have a `.code` though, so let's remove that? Or am I missing something?
```suggestion
            if not isinstance(err, SystemExit):
                error_handler(cls, err, tb, self._action)
                sys.exit(1)
```

bdieterm: I guess you will surely already know the answer:
The expression on the right of the `or` will only be evaluated if the left part equals `False`, because of the short circuit evaluation.
Therefore `err.code` is only accessed if `err` is an instance of `SystemExit`.

almarklein: Ah yeah, I missed that it'd short-circuit. Good call to cover this case 👍

PR #862, opened 2023-05-17
fix QFontDatabase compatibility for PyQt6

bdieterm:
When running Pyzo in PyQt6 the menu "View --> Font" did not work because PyQt6 does not allow `QtGui.QFontDatabase()` anymore (but PySide6 still does).


almarklein:
🚀

PR #868, opened 2023-05-29
improve PDF export

bdieterm:
Features of the new version:
* significant speed improvements for preview/export of larger documents (e.g. 6500 lines with line numbers is down from 130 seconds to a few seconds)
* parser for syntax highlighting is now the same as in the editor window (instead of Manager.suggestParser)
* added missing line that I accidentally deleted in my last pull request
* added "Done" button in the dialog window
* added proper error message box when the PDF could not be saved
* zoom level slider offset removed (0 is normal, negative values are smaller font sizes)
* added workaround for wrong preview in Qt6 when changing from portrait to landscape
* updated code style to better match the rest of the code base, and cleaned up the code a bit

### DIFF
--- a/pyzo/core/pdfExport.py
+++ b/pyzo/core/pdfExport.py
@@ -1,4 +1,5 @@
 import os
+import sys
 import pyzo
 from pyzo import translate
 from pyzo.qt import QtCore, QtGui, QtWidgets, QtPrintSupport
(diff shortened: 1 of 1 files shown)


almarklein:
Those are nice changes! 

> significant speed improvements for preview/export of larger documents (e.g. 6500 lines with line numbers is down from 130 seconds to a few seconds)

Could you explain how you managed to do this? 🚀

PR #876, opened 2023-06-01
fix truncated version info string in pyzo logger welcome message

bdieterm:
Without this patch, the Pyzo Logger displays for example
`Python 3.11. on linux - This is the Pyzo logger shell.`
instead of
`Python 3.11.0 on linux - This is the Pyzo logger shell.`.


almarklein:
Oh wow, that code is so old that it assumed the `x` in Python version `3.x` to be `<=9` 😄

Issue #698, opened 2020-03-24
Settings -> Edit syntax styles... -> New...  crashes


ducciov:
with pyzo-4.10.2.2-linux64.tar.gz for debian
(pyzo 4.9.0 flatpak do not have this problem)

If I go to main menu "Settings -> Edit syntax styles..."
and in that window I choose "New" and than in the same combo I select any other predefined style ("default", "dark", "scintilla"....) Pyzo closes without saving
pending changes to file.

If I select new style and than "Save" and "Apply theme" Pyzo do not crash
but than I can never no more change to preset style becouse if I try Pyzo
crashes. I restart Pizo but always the same behavior. The only workaround
I have found is "Settings" -> "Advanced Settings..." and than select button
"Factory Default".

almarklein:
I cannot reproduce this, tried on Windows and Ubuntu 18.04. What I tried:
* Create new style
* Save the style
* Select dark style

Note that when you hit "apply", Pyzo restarts to apply it. It may be better communicated that this involves a restart.

ducciov:
The crash happens without "save" or "apply",
In the window "Edit syntax styles..." in the combobox of styles I just click "New" and then click "scintilla": Pyzo immediately closes (doesn't restart).

Maybe I am wrong as I have not installed.
I have downloaded and extracted pyzo-4.10.2.2-linux64.tar.gz.
Then I start Pyzo with:
>> python3 __main__.py

almarklein:
Wait, what? If you downloaded the tarfile, you can (and *should*) just run `./pyzo` :)

Tontyna:
@almarklein Provided that PyQt5 is installed in the python3 environment you can start Pyzo with

```
python3 __main__.py
```

No problem. You implemented it:

https://github.com/pyzo/pyzo/blob/39dec5c74c1c83190533d8358559f2015c30d54c/pyzo/__main__.py#L8-L11

Tontyna:
When I try to 'Edit syntax styles' in Pyzo 4.10.2 I get the following message on stdout:

```
Exception while setting style syntax.illegal for theme default : 'syntax.illegal'
```

The 'real' error is a *KeyError* at line 394, in setTheme()

https://github.com/pyzo/pyzo/blob/39dec5c74c1c83190533d8358559f2015c30d54c/pyzo/core/themeEdit.py#L392-L396

[...]

PR #878, opened 2023-06-02
add missing style entry for Syntax.illegal

bdieterm:
In the syntax/theme editor, style entry "Syntax.illegal" the color entry widgets are all empty for all themes (including the default theme), because in the *.theme files, the entry "syntax_illegal" was missing.

This missing entry has already been mentioned some time ago in #698.

This PR adds the missing lines in the themes files.


almarklein:
Nice find!

PR #879, opened 2023-06-11
save "Auto hide" value in editor search dropbox

bdieterm:
The value for the "Auto hide" checkbox is read from the config file but was never saved back to it unlike the other checkboxes.
With this PR, the state of the "Auto hide" checkbox will be saved.


almarklein:
Thanks!

PR #880, opened 2023-06-11
reduce search dropbox margins

bdieterm:
In my opinion, the borders around the search dropbox are a waste of space.
With this PR, I can see ca. 1.5 text lines more in the editor when the search dropbox is visible.


almarklein:
Seems fair.

PR #881, opened 2023-06-15
python 2.7 compatibility

bdieterm:
I just stumbled upon this:
Pyzo still supports Python 2.7 shells/kernels (after correcting one line with this PR).

@almarklein:
In my opinion we could drop Python 2.7 support (as well as Qt4 support) to modernize the codebase.
What do you think about that?

### DIFF
--- a/pyzo/pyzokernel/guiintegration.py
+++ b/pyzo/pyzokernel/guiintegration.py
@@ -115,7 +115,7 @@ class App_nogui(App_base):
 
     def run(self, repl_callback):
         # Move at a slow pace - there is no gui to keep running
-        super().run(repl_callback, 0.1)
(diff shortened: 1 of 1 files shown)


almarklein:
Up until recently (it might still be) Pyzo could work with kernels from Python 2.4 and up (the version in use when I started Pyzo) 😆 

I think its a very powerful idea that you can just run Pyzo and it can fire up any Python interpreter on the system. Being able to launch 2.4 is a bit silly at this point, but 2.7 is still installed on a lot of systems. So I'd prefer to keep the kernel code working on 2.7 for a while longer. That does not mean we should put a lot of maintenance effort into keeping all GUI systems working on 2.7 as well. If these fail on 2.7 I don't think many people would blame us :)

PR #882, opened 2023-06-16
add new tool "EditorList"

bdieterm:
When there are many tabs open in the editor, switching between them gets cumbersome. So I created a new tool:

The new EditorList is similar to the editor's tab widget.

Mouse Control
* right click on a list entry opens the context menu (the same as the tab's)
* double click on a list entry activates the corresponding editor tab (this is different, the tab saves on double click)
* double click below the last list entry opens a new editor tab
* middle click on a list entry closes the corresponding editor tab
* drag and drop to sort list entries, synchronized with the editor tabs

Keyboard Control
* ENTER or RETURN activates the selected editor tab
* DELETE will close the selected editor tab
* INSERT opens a new editor tab
* CTRL+SHIFT + UP/DOWN shifts the entry up/down (and the editor tab left/right)

There was also a bug in editorTabs.py that I fixed:
When starting Pyzo with just a single file open that is pinned and/or set as main file, that will not be updated till updateItems() is executed (e.g. when creating a new file).

### DIFF
--- a/pyzo/core/editorTabs.py
+++ b/pyzo/core/editorTabs.py
@@ -1542,6 +1542,8 @@ class EditorTabs(QtWidgets.QWidget):
                 self.newFile()
         else:
             self.newFile()
+        self._tabs.updateItems()  # without this, a single restored editor will not have
+        # correct PINNED/MAINFILE symbols
 
         # The find/replace state is set in the corresponding class during init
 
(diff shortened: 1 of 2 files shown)


almarklein:
This already looks pretty nice, I think it could use some styling to make it look more like tabs, but I can see how it can already be useful. So I have no objections to adding it right now.

almarklein:
I plan on doing a release this weekend, perhaps tonight.

almarklein:
> I plan on doing a release this weekend, perhaps tonight.

CI is stuck on outdated certs for signing the macos app bundle. Well look into it later this weekend.

PR #883, opened 2023-06-18
replace version string parser with own implementation

bdieterm:
Currently, there are two use cases for converting a string, like "123.45ew6.7x.dev8", to a comparable tuple, like (123, 45, 7).

1. `pyzo/pyzokernel/interpreter.py`
The version comparison is just to check if the macOS version is >= 10.9.
Before this PR, `distutils.version.LooseVersion` was used, but `distutils` is not available anymore in Python 3.12. See [PEP 632](https://peps.python.org/pep-0632/) and the [release notes](https://www.python.org/downloads/release/python-3120b2/) for details.

2. `pyzo/qt/__init__.py`
The version comparison is to check macOS and Qt versions.
Before this PR, the external package `packaging` was used for this (and only for this).

With the own implementation, pyzo has no more external dependencies, except `dialite` and a Qt binding.

The new function `parse_version_crudely` is implemented twice because in case (1) it is executed by the shell/kernel's Python interpreter and in case (2) by the Python interpreter that runs the Pyzo IDE. I think code duplication does not harm here because the function is merely one line of trivial code. And before this PR, two different packages (distutils and packaging) have been used for the same task. Using a common code file would be possible: this would require a new folder such as "kernel_and_ide" to have a clean separation. Currently all files in directory `pyzo/yoton` are the only code files that are used by both the kernel and the IDE.

### DIFF
--- a/freeze/frozen_libs.txt
+++ b/freeze/frozen_libs.txt
@@ -1,2 +1,1 @@
-packaging
 dialite
--- a/pyzo/pyzokernel/interpreter.py
+++ b/pyzo/pyzokernel/interpreter.py
@@ -35,7 +35,6 @@ import traceback
 import keyword
 import inspect  # noqa - Must be in this namespace
 import bdb
-from distutils.version import LooseVersion as LV
 import linecache
 
 import yoton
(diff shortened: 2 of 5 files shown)


almarklein:
Nice, one dependency less :)

PR #886, opened 2023-06-20
fix appconfig path on MS Windows and clean up code

bdieterm:
### Description of the problem
I encountered the following problem on different computers with MS Windows 10:
After installing and running Pyzo the first time, the user confirms the auto-detected pyzo interpreter or selects an individual one. But when closing Pyzo and starting it again, Pyzo does not remember the previous settings and the user has to configure everything again. After the second time, saving and loading the configuration works normally.
### Why does this happen?
When starting Pyzo for the first time (i.e. no previously saved configurations found), Pyzo selects the following paths for the appdata and config by executing function `appdata_dir` in `pyzo/util/paths.py`. The results are:
```
[Logger] >>> pyzo.appDataDir, pyzo.appConfigDir
('C:/Users/user/AppData/Roaming\\pyzo', 'C:/Users/user/AppData/Local\\pyzo')
```
When starting Pyzo the first time, the folder `C:\Users\user\AppData\Roaming\pyzo` (with `command_history.py`, etc.) is created, and when closing Pyzo the first time, settings are saved to file `C:\Users\user\AppData\Local\pyzo\config.ssdf`.

When starting Pyzo the second time (and every time after that), function `appdata_dir` behaves differently because folder `C:\Users\user\AppData\Roaming\pyzo` already exists:
```
[Logger] >>> pyzo.appDataDir, pyzo.appConfigDir
('C:\\Users\\user\\AppData\\Roaming\\pyzo', 'C:\\Users\\user\\AppData\\Roaming\\pyzo')
```
So, the second time (and later times) Pyzo is started, Pyzo will use the appdata path also as the config path. And therefore it ignores the file `C:\Users\user\AppData\Local\pyzo\config.ssdf` from the very first time that Pyzo was run. The config will be loaded and saved from/to `C:\Users\user\AppData\Roaming\pyzo\config.ssdf` instead from the second time on.

### Solution
I decided to keep the common directory for appdata and config (on MS Windows operating systems). This will keep the newest configuration on existing Pyzo installations after updating Pyzo. And I also cleaned up the code -- there were even remanents from the old times when Pyzo was a full Python distribution with normal Python interpreters that could be used by a shell/kernel in the Pyzo IDE.

I tested the changes on MS Windows 10 and Linux, and also the portable settings directory. But I do not have the equipment to check if Pyzo behaves still the same on macOS.

### DIFF
--- a/pyzo/_start.py
+++ b/pyzo/_start.py
@@ -102,28 +102,6 @@ sys.excepthook = pyzo_excepthook
 
 # todo: move some stuff out of this module ...
 
-
-def getResourceDirs():
-    """getResourceDirs()
-    Get the directories to the resources: (pyzoDir, appDataDir, appConfigDir).
-    Also makes sure that the appDataDir has a "tools" directory and
-    a style file.
-    """
-
-    pyzoDir = os.path.abspath(os.path.dirname(__file__))
-    if ".zip" in pyzoDir:
-        raise RuntimeError("The Pyzo package cannot be run from a zipfile.")
-
-    # Get where the application data is stored (use old behavior on Mac)
-    appDataDir, appConfigDir = paths.appdata_dir("pyzo", roaming=True, macAsLinux=True)
-
-    # Create tooldir if necessary
-    toolDir = os.path.join(appDataDir, "tools")
-    os.makedirs(toolDir, exist_ok=True)
-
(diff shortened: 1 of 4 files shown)


almarklein:
Wow, I can imagine this took some digging ... All looks fine on MacOS!

PR #895, opened 2023-09-17
fix daemon threads

bdieterm:
Changes:

- in kernelbroker.py there was a spelling mistake in `daemon` that created an unused attribute instead of setting the property
- `setDaemon()` is deprecated and therefore replaced by setting the property `daemon` depending on the interpreter version

### DIFF
--- a/pyzo/core/kernelbroker.py
+++ b/pyzo/core/kernelbroker.py
@@ -743,7 +743,7 @@ class StreamReader(threading.Thread):
         self._process = process
         self._strm_raw = strm_raw
         self._strm_broker = strm_broker
-        self.deamon = True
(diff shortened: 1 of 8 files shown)


pyzo/core/kernelbroker.py
```diff
@@ -743,7 +743,7 @@ def __init__(self, process, strm_raw, strm_broker):
         self._process = process
         self._strm_raw = strm_raw
         self._strm_broker = strm_broker
-        self.deamon = True
+        self.daemon = True
```
almarklein, line 746: This is exactly why I preferred the method 😉

almarklein (approved):

PR #896, opened 2023-09-17
upgrade from deprecated imp module to importlib

bdieterm:
Python 3.12 has no more `imp` module. Therefore I replaced it with `importlib.util`.


almarklein:
Nice!

PR #897, opened 2023-09-17
fix and clean regular expressions

bdieterm:
When executing Pyzo from source with env variable `PYTHONWARNINGS=error` I got some errors about improper escaped strings in regular expressions. I fixed these and also cleaned up other re strings.


almarklein:
LGTM, thanks!

PR #898, opened 2023-09-17
use context manager for file access

bdieterm:
When executing Pyzo from source with env variable `PYTHONWARNINGS=error` I got many warnings because of files that were not closed (immediately). I changed the code to use the context manager whereever suitable, and I also cleaned up some mode and encoding arguments.

### DIFF
--- a/freeze/pyzo_freeze.py
+++ b/freeze/pyzo_freeze.py
@@ -117,8 +117,8 @@ excludes = []
 includes += get_stdlib_modules()
 
 # Include a few 3d party packages, e.g. deps of qtpy
-with open(os.path.join(this_dir, "frozen_libs.txt"), "rt") as fd:
-    includes += fd.read().split()
+with open(os.path.join(this_dir, "frozen_libs.txt"), "rb") as fd:
+    includes += fd.read().decode().split()
 
 # Include a subset of Qt modules
 qt_includes = [
--- a/freeze/pyzo_freeze.py
+++ b/freeze/pyzo_freeze.py
@@ -30,7 +30,9 @@ def get_pyzo_version():
     """Get Pyzo's version."""
     filename = os.path.join(dist_dir, "..", "..", "pyzo", "__init__.py")
     NS = {}
-    for line in open(filename, "rb").read().decode().splitlines():
+    with open(filename, "rb").read() as fd:
+        data = fd.read()
+    for line in data.decode().splitlines():
         if line.startswith("__version__"):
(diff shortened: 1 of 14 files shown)


pyzo/tools/pyzoFileBrowser/proxies.py
```diff
@@ -416,7 +416,9 @@ def fileSize(self, path):
 
     def read(self, path):
         if op.isfile(path):
-            return open(path, "rb").read()
+            with open(path, "rb") as f:
+                data = f.read()
+            return data
```
almarklein, line 421:
```suggestion
                return f.read()
```

pyzo/util/interpreters/__init__.py
```diff
@@ -175,11 +175,12 @@ def _get_interpreters_conda():
     exes = []
     filename = os.path.expanduser("~/.conda/environments.txt")
     if os.path.isfile(filename):
-        for line in open(filename, "rt").readlines():
-            line = line.strip()
-            exe_filename = os.path.join(line, pythonname)
-            if line and os.path.isfile(exe_filename):
-                exes.append(exe_filename)
+        with open(filename, "rt") as fd:
```
almarklein, line 178:
```suggestion
        with open(filename, "rt", encoding="utf-8") as fd:
```

freeze/pyzo_freeze.py
```diff
@@ -115,7 +117,8 @@ def get_stdlib_modules():
 includes += get_stdlib_modules()
 
 # Include a few 3d party packages, e.g. deps of qtpy
-includes += open(os.path.join(this_dir, "frozen_libs.txt"), "rt").read().split()
+with open(os.path.join(this_dir, "frozen_libs.txt"), "rt") as fd:
+    includes += fd.read().split()
```
almarklein, line 121:
```suggestion
with open(os.path.join(this_dir, "frozen_libs.txt"), "rb") as fd:
    includes += fd.read().decode().split()
```

almarklein (commented):
Nice work, really appreciate it!

I added just a few suggestions to consistently use utf-8.

pyzo/util/interpreters/__init__.py
bdieterm: Good catch! I was not sure about the encoding used by conda's environments.txt so I did not touch this.
Looking into that it turns out that on older conda versions the native text encoding of the operating system was used and in newer conda versions utf-8, see https://github.com/conda/conda/issues/8662

PR #900, opened 2023-09-17
replace deprecated distutils package

bdieterm:
The `distutils` package was removed in Python 3.12. I created a function that resembles the important part of it without external dependencies.

### DIFF
--- a/freeze/pyzo_freeze.py
+++ b/freeze/pyzo_freeze.py
@@ -74,6 +74,7 @@ def _find_modules(root, extensions, skip, parent=""):
 
 
 def get_python_stdlib_path():
+    # This code is extracted from distutils.sysconfig.get_python_lib (distutils is deprecated in 3.12)
     stdlib_path = None
     prefix = os.path.normpath(sys.base_prefix)
     if os.name == "posix":
--- a/freeze/pyzo_freeze.py
+++ b/freeze/pyzo_freeze.py
@@ -6,7 +6,6 @@
 import os
 import sys
 import shutil
-from distutils.sysconfig import get_python_lib
 
 
 # Definitions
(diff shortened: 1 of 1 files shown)


freeze/pyzo_freeze.py
```diff
@@ -74,9 +73,23 @@ def _find_modules(root, extensions, skip, parent=""):
                     yield moduleName
 
 
+def get_python_stdlib_path():
```
almarklein, line 76: So if I understand correctly, the code below is copied (more or less) from `distutils`? It feels kind of fragile to have this triage here like that. Knowing what the original source was gives some reassurance 😄 so let's add a comment if that's indeed the case.
```suggestion
def get_python_stdlib_path():
    # This code is extracted from distutils.sysconfig.get_python_lib (distutils is deprecated in 3.12)
```

almarklein:
Thanks for the amazing work @bdieterm! Apart from the improvements being really substantial, the PR's themselves are also a joy to review 😃

freeze/pyzo_freeze.py
bdieterm: The original implementation can be seen here:
https://github.com/python/cpython/blob/a673248d6cb1ae5d5b7fbf0ab6c8db8c32df0eee/Lib/distutils/sysconfig.py#L304-L346

bdieterm:
> Thanks for all the amazing work @bdieterm! Apart from the improvements being really substantial, the PR's themselves are also a joy to review 😃

Happy to read that 😃, you are welcome.

BTW, thanks for creating Pyzo, @almarklein, and tanks to all contributors!
For my use cases, mostly scripts and smaller engineering related projects, Pyzo is a real joy to use.
I am amazed by the software architecture and by the many thoughtful little details and features that went into the code.

PR #904, opened 2023-09-22
use fontWeight enum names instead of integers

bdieterm:
From Qt5 to Qt6 the numeric values for `setFontWeight` changed:
Qt5 ... light: 25, normal: 50, bold: 75 ... see https://doc.qt.io/qt-5/qfont.html#Weight-enum
Qt6 ... light: 300, normal: 500, bold: 700 ... see https://doc.qt.io/qt-6/qfont.html#Weight-enum

To avoid Qt-version dependent value pairs I changed the code to use enum names (e.g. `QtGui.QFont.Bold`) instead of integers.


almarklein:
Wow ....

PR #906, opened 2023-09-22
fix shell prompt for debugging with stack frames

bdieterm:
When running a script that makes use of the Qt event loop and also using breakpoints, the breakpoints could be triggered by either
- the normal linear execution (while the shell is busy), or
- code executed in the event loop.

In the latter case, i.e. setting a breakpoint in a callback function, the shell's prompt does not change to the stack frame when triggering the breakpoint. One has to debug-step or press the RETURN key to update to the correct prompt text. The same happens, when leaving the debug mode.

While fixing this, I encountered a bug that caused wrong formatting, and fixed it -- see #905.

For the sake of completeness, this is the code I used to reproduce the problem and test the new behaviour:
```
from PySide2 import QtGui, QtCore, QtWidgets

class Example(QtWidgets.QPushButton):
    def __init__(self):
        super().__init__()
        self.setFixedSize(300, 100)
        self.setText('click me')
        self.clicked.connect(self.callback_button)
        self.show()

    def callback_button(self):
        print('callback_button')
        func2()  # call another function just to have multiple stack frame levels


def func2():
    a = 1
    b = 2  # <-- set a breakpoint at this line
    c = 3
    d = 4


example = Example()
```

### DIFF
--- a/pyzo/core/shell.py
+++ b/pyzo/core/shell.py
@@ -818,8 +818,9 @@ class BaseShell(BaseTextCtrl):
             self._cursor1.removeSelectedText()
             self._cursor2.setPosition(self._cursor1.position(), A_MOVE)
         elif prompt == 2:
-            # Insert text after prompt, inserted text becomes new prompt
-            self._cursor1.setPosition(self._cursor2.position(), A_MOVE)
+            # text becomes new prompt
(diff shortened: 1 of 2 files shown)


almarklein:
You dig deep!

PR #910, opened 2023-10-01
improve shell printing performance for longer texts

bdieterm:
This PR improves the performance of printing text in the shell.
Especially when printing many short lines, e.g. 10 characters per line, this results in a 5x speed-up.
```
import timeit
t = timeit.timeit("print(('a'*10 + '\\n')*20000)", number=10)
```
As a positive side-effect I also observed that the editor has a much better responsiveness. While the code that prints the text is (still) running, quickly typed characters do not show up immediately without this patch.


almarklein:
🚀  sweet!

PR #912, opened 2023-10-13
activate breakpoint() only for Python >= v3.7

bdieterm:
When I added the support for the built-in function `breakpoint()` (see #908), I forgot to only activate this for newer Python versions. Consequently, shells with Python versions < 3.7 crashed during start-up because of the missing `breakpointhook` attribute in the `sys` module. This PR will fix the problem by only activating `breakpoint()` for newer Python versions.

### DIFF
--- a/pyzo/pyzokernel/debug.py
+++ b/pyzo/pyzokernel/debug.py
@@ -19,7 +19,7 @@ class Debugger(bdb.Bdb):
         bdb.Bdb.__init__(self)
         self._debugmode = 0  # 0: no debug,  1: postmortem,  2: full debug
         self._files_with_offset = []
-        if sys.version_info >= (3, 7):
+        if hasattr(sys, "breakpointhook"):
             self._original_breakpointhook = sys.breakpointhook
             sys.breakpointhook = self.custom_breakpointhook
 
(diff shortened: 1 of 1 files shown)


bdieterm:
Hi @almarklein!
I am wondering why this PR has not been merged.
Without the proposed changes, Pyzo cannot start Python shells with Python version < 3.7 anymore. E.g. when opening an old Python 2.7 shell (only for testing), the Pyzo shell widget will only display `The kernel process exited. (1)`.

pyzo/pyzokernel/debug.py
```diff
@@ -19,8 +19,9 @@ def __init__(self):
         bdb.Bdb.__init__(self)
         self._debugmode = 0  # 0: no debug,  1: postmortem,  2: full debug
         self._files_with_offset = []
-        self._original_breakpointhook = sys.breakpointhook
-        sys.breakpointhook = self.custom_breakpointhook
+        if sys.version_info >= (3, 7):
```
almarklein, line 22: Ducktyping this, because I'm a bit unsure about other interpreters (e.g. pypy) implementhing this.
```suggestion
        if hasattr(sys, "breakpointhook"):
```

almarklein:
My bad, this slipped my attention.

Issue #915, opened 2023-11-17
Dark mode file tabs barely readible


TeunBartelds:
Decided to create a separate issue for this specific problem, however it is possibly related to #914.

I really like the dark mode, but cannot easily read the file names of open files (using Windows 10 on dark mode and Pyzo 4.13.3):
![image](https://github.com/pyzo/pyzo/assets/11867578/e1dcad10-b436-4319-bc53-f29ef51dc5c1)

Is it possible to change the font and/or background colour of the filename tabs? Either as a setting or in the source code?

PR #916, opened 2023-11-17
disable custom text colors for editor tabs

bdieterm:
As described in #915, the dark text in the editor tabs is barely readable in Qt 6.5 dark mode.

I actually had to look twice to realize that Pyzo sets different text colors for the active editor tab and for the main file editor tab. Therefore I suggest just disabling these custom text colors until a proper dark mode handling is implemented.
The main file tab is annotated with a star symbol anyways, an the active tab also has a different shape, so the differentiation is still possible.

After this fix, the editor tab texts will have a proper text color, i.e. black text in light mode an white text in dark mode.

### DIFF
--- a/pyzo/core/editorTabs.py
+++ b/pyzo/core/editorTabs.py
@@ -931,13 +931,16 @@ class FileTabWidget(CompactTabWidget):
             else:
                 tabBar.setTabToolTip(i, item.filename)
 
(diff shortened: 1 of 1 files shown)


bdieterm:
I already submitted a new PR with proper color selection depending on the light/dark mode.

Issue #804, opened 2022-03-26
Autoindent does not work if line ends with comment


ysalmon:
I just realised that Pyzo does not auto-indent the following line on pressing enter if the previous line end with a comment :

```python
def f(x) :
   x = x + 1    # this line is autoindented
   if x > 0 :  # because of this comment, the follwing line will not be auto-indented
      #TODO manually type Tab to indent !
```

bdieterm:
@almarklein: this can be closed (already fixed)

PR #920, opened 2023-11-19
improve python auto-indent

bdieterm:
This PR will make the editor's auto-indent also work for comments after the ":", e.g.:
```
if x > 4:  # check size [PRESS RETURN TO GET TO NEXT LINE]
    y = ...  # this line is automatically indented correctly
```
This will solve #804.


almarklein:
I don't miss this a lot, but when it *does* happen, I always cringe 😆 , so it's really nice to see this added! 🚀

almarklein (approved):

PR #931, opened 2023-11-28
fix filebrowser tool namefilter combination

bdieterm:
When setting the filename filter in the "File Browser" tool to a suggested default value, e.g. `*.py *.pyw`, then no more *.py files are visible because in the new implementation, the filename only matches if all filter parts match (logical AND).
I reverted parts of #926 to get back the old behaviour (logical OR).


almarklein:
I apologize, I rushed that PR a bit and if I had only looked at the preset filters I would have seen why the filter was OR. Thanks for the fix 😊

PR #936, opened 2023-12-01
add menu entry for search in file browser tool

bdieterm:
This PR adds a menu entry for searching via the File Browser tool.
Similar to the normal search (Ctrl+F), the selected text in the editor is copied into the search field and the focus is moved to the search field.
This also works if no editor tab is open.
Default keyboard shortcut is Ctrl+Shift+F.

### DIFF
--- a/pyzo/core/menu.py
+++ b/pyzo/core/menu.py
@@ -819,12 +819,30 @@ class EditMenu(Menu):
             None,
             pyzo.editors._findReplace.findPrevious,
         )
+        self.addItem(
+            translate(
+                "menu",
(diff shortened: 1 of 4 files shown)


bdieterm:
The two failed checks seem to be an issue with the newest PyQt6 on Py3.11 on Linux and MS Windows.
Installing PyQt6==6.6.0 in a Python 3.11 on my Linux computer also does not work -- there is an error when running `from PyQt6 import QtCore`.

almarklein (approved):
LGTM.

Let's assume (for now) that these errors will be fixed upstream. If not we'll look into them at another time.

Issue #875, opened 2023-05-31
Windows file dialog causes errors after using QFileIconProvider


bdieterm:
When starting Pyzo in Windows via pyzolauncher.py the native file dialog (open/save) will raise errors
`Windows fatal exception: code 0x8001010e` under certain conditions.
To see these errors, the Python `faulthandler` has to be enabled, as it is done in pyzolauncher.py.
The Pyzo tool "File Browser" has to be open in Pyzo and it has to show an icon extracted from a file (but not a folder), for example the symbol of a *.py file, because this will call a function from QFileIconProvider.
Opening a native file dialog in Pyzo (no matter if opening/closing a file in the editor or exporting a PDF) will then cause these errors in the background, but they seem to be in another thread and Pyzo does not seem to be affected by this.

When starting Pyzo with the "File Browser" tool closed and then opening a file dialog, there are no errors. And after this, the File Browser tool can be opened and this will not cause errors anymore.

I narrowed down the problem and recently [posted it to this Qt forum](https://forum.qt.io/topic/145439/windows-file-dialog-causes-errors-after-using-qfileiconprovider). So this seems to be more a Windows/Qt issue than a Pyzo issue.

If someone wants to avoid these errors, these workarounds are currently possible in Pyzo:
- when running Pyzo from source, `import pyzo; pyzo.start()` instead of running pyzolauncher.py, the application will not flood the console with these errors but hide them instead
- using non-native file dialogs, as implemented in #873, will prevent the errors
- using `iconprovider.icon(iconprovider.File)` instead of `iconprovider.icon(QtCore.QFileInfo(dummy_filename))` in pyzo/tools/pyzoFileBrowser/tree.py will prevent the errors by using a generic file icon
- not using the Pyzo "File Browser" tool or opening it after first opening a file dialog will also prevent the errors

@almarklein 
I think this issue is not critical and we have workarounds, so now would be a good time for a new Pyzo release.

almarklein:
Closed by #937

PR #937, opened 2023-12-02
disable faulthandler in pyzolauncher.py in MS Windows to prevent crash

bdieterm:
This PR will change pyzolauncher.py so that module "faulthandler" will not be used on MS Windows operating systems. This will prevent crashes when using the "File Browser" tool and native file dialogs when Pyzo is launched via pyzolauncher.py.
See issue #875 for details.

Deactivating the faulthandler on MS Windows is acceptable IMHO. When running Pyzo via the binary distribution (pyzo.exe) or via `python -m pyzo` there has also no faulthandler been activated.

### DIFF
--- a/pyzolauncher.py
+++ b/pyzolauncher.py
@@ -29,6 +29,9 @@ import subprocess
 try:
     if sys.executable.lower().endswith("pythonw.exe"):
         raise ImportError("Dont use faulthandler in pythonw.exe")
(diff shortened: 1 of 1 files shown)


almarklein (approved):

almarklein:
If this fixes the crashing, it's well worth turning it off 👍

PR #943, opened 2023-12-12
Fix CI build (for macos, py312, pyside6)

almarklein:


almarklein:
image

So it looks like PySide6 is at fault (not the combination with py312). In particular the latest version 6.6.1.

almarklein:
For what it's worth. Pyside 6.6.1 works fine on my Mac locally. So maybe the fact that this runs on CI is also a factor. Anyway, I also pinned the building of binaries to 6.6.0 for Mac, for now.

Issue #891, opened 2023-07-02
Autocompletion for variables not available


UniqASL:
Hi,

I update pyzo on Windows 10, from 4.12.3 to the latest 4.13.3 version and somehow the environment variables are not available anymore (before they were suggested after writing the first 3 letters). Functions are still available but not single variables.

**Version info**
Pyzo version: 4.13.3 (binary)
Platform: win32
Python version: 3.10.11
Qt version: 6.5.1
PySide6 version: 6.5.1.1

almarklein:
Thanks for the report. I just experimented a bit. When code has been executed, and the variables exist in the shell, then all is right. When the variables only live in the editor yet, they don't autocomplete... but this seems to have been so for a while (last release that I tried was 4.12.3). Could you please be a bit more specific, as to what you do, what you expect to happen, and what happens instead?

UniqASL:
Thanks for the reply. I made some more test and I found out that it only happened with one script. The issue was that I was using the variable `dir` in this script, which I should'nt have. No that I renamed the variable, it works. Sorry it had nothing to do with Pyzo.

UniqASL:
Just a thought. To prevent that it would be nice if built-in names were shown in a different color / style by default. I found out how to change that in the style editor, but having it as default would be an advantage (most people probably don't parametrize the styles themselves).

almarklein:
Closing this, as its possible to edit the syntax style, and making all builtin a different color by default is a pretty big impact (since there are a lot of them) and I don't think everybody wants that.

PR #947, opened 2024-01-04
always use built-in dir command for autocompletion introspection

bdieterm:
As described in issue #891, naming a variable "dir" will break autocompletion in Pyzo.

With this PR, the introspection for autocompletion will always use the built-in "dir" command, even when the user executed code such as `dir = 'northeast'`.


almarklein (approved):
Smart!

PR #949, opened 2024-01-18
avoid crash caused by faulty QTextBlock

bdieterm:
It already happened to me two or three times that Pyzo suddenly crashed when doing many undo-operations in close succession (having keys Ctrl+Z pressed for several seconds).

I found a way to reproduce such a crash:
Start Pyzo and execute the following code in the Logger tool:
```Python3
exec(r"""
editor = pyzo.editors.newFile().editor
for _ in range(50):
    editor.insertPlainText(('a'*20 + '\n')*20)
editor.toggleBreakpoint(20)
for _ in range(49):
    editor.undo()
# editor.undo()  # --> this (or Ctrl+Z) will crash Pyzo
""")
```
(The exec with the multiline string is necessary because only one command can be entered and executed at a time in the Logger shell.)
This will not yet crash Pyzo because the last line in the code is deactivated by a comment sign. To finally crash Pyzo, just do a single "undo" operation (Ctrl+Z or "Edit -> Undo" in the menu).
On my Linux system, this crashed Pyzo every time I did this.
On a MS Windows system, the crash mostly happend after trying this 1 to 5 times.

So the crash happens on both Linux and MS Windows and with all Qt wrappers I tested (Qt5/6 and PySide/PyQt), even when running Pyzo from source.

The error message after the crash is "Fatal Python error: Segmentation fault", and according to the stack trace the line `prev_ok = block.previous().blockNumber() == block_previous.blockNumber()` in file pyzo/codeeditor/extensions/appearance.py was the last executed python code.

I have seen that the crash happened after executing certain methods of the previous QTextBlock object, such as `block.previous().blockNumber()` or `block.previous().firstLineNumber()`. It seems that the memory of the "previous" block was already freed by a bug in the Qt library.
Luckily some methods still work without crashing and give very unusual values when dealing with a faulty block, so I added code to check for that and ignore the blocks.

This is just a workaround to prevent Pyzo from crashing and losing unsaved data.

@almarklein: Could you please test reproducing the error on a macOS computer and check if the workaround works there as well?


almarklein:
I tried on MacOS 14.3 (M1 pro) with PySide 6.6.0 and Python 3.11.2. I tried 3 times:
* Crashed after the second attempt.
* Crashed at the first attempt.
* Crashed at the 5th or so attempt.

The error is just
```
[1]    33797 bus error  pyzo
```

Trying again, but with the change proposed in this PR, I get the "prevented segfault" warning about half the time, but have not seen a crash after 10 tries or so.

almarklein (approved):
Great work tracing this down! Unfortunate that we cannot actually fix it, but this will sure avoid frustration for many! 👍

almarklein:
Ready from your end?

bdieterm:
I did further tests with the previous workaround. One time a crash happend while working normally but I could not reproduce that.
So I looked for better ways to fix that problem. My first idea was to add a `userData` object that has a destructor callback (`def __del__(self):`) to the tracked blocks but it turned out that the object was not destroyed for a faulty object before the crash.

I finally came up with the following approach:
Instead of checking if `block.previous()` is a faulty block, I check if the `block` that is assigned to the breakpoint was restored by Qt during an undo operation. In that case the breakpoint will then be deleted.

I think that this new approach will work much better to prevent crashes and that the code is cleaner.
This PR is now ready to be merged into main.

<details>
<summary>Just for the record: my first commit (that was overwritten by the force-push)</summary>

code added between lines 1212 and 1213 in appearance.py
```python3
            # Apparently there is a bug in Qt5 and Qt6 that can cause a segmentation fault.
            # When there is a faulty block "bp", calling methods such as bp.blockNumber()
            # will crash Pyzo immediately. Luckily some methods give very unusual values
            # when dealing with a faulty block. See https://github.com/pyzo/pyzo/pull/949
            bp = block.previous()
            if (
                not (1 <= bp.length() < 1000)
                or bp.userState() == bp.revision() == bp.lineCount()
            ):
                print(
                    "WARNING: avoiding segfault in ",
                    self.__class__,
                    "__onBlockCountChanged",
                )
                continue
```
</details>

bdieterm:
I found out that the faulty blocks can also occur without undo operations.
And I also found a very simple way to cause such a crash with the current Pyzo release (before this PR):
```
Ctrl+N ... create a new file 
RETURN ... new line
RETURN ... new line
RETURN ... new line
Ctrl+B ... set breakpoint in line 4
RETURN ... new line
Backspace ... delete current empty line
--> breakpoint unexpectedly jumps from line 4 to line 2
Ctrl+A ... select all
DEL ... delete selection
--> Pyzo crashes
```
My previous commit also correctly avoided a crash for that situation. With the new force-pushed commit I additionally updated the comments and refined the code.

almarklein (approved):
Nice work @bdieterm !

PR #954, opened 2024-02-03
Fix unused default settings

bdieterm:
This PR will remove two unused settings in defaultConfig.ssdf: "autoComplete_fillups" and "homeAndEndWorkOnDisplayedLine".
Both entries have never been used as far as the commit history goes back here on GitHub.

And for the third entry, that has also not been used, "autoComplete_caseSensitive", I have added code to implement that feature.


almarklein:
I suppose the reason/implementation is lost. Years ago this IDE was called IEP. Not sure whether I created a new repo when I renamed it. But it might also be related to using `main` branch instead of `master`.  Anyway, thanks for the fix! 😆

PR #955, opened 2024-02-10
add breakpoint disabling and breakpoint navigation

bdieterm:
Motivation for these changes was to have "navigation points" within a file for easier jumping around vertically between marked rows. In order to keep things simple, the breakpoints in the editor will double as the row markers. And if a breakpoint should only be a navigation marker, it can be toggled to a "disabled breakpoint".

So the first part is to have tri-state breakpoints, which toggle in the following sequence:
removed (no breakpoint) -> on (whole filled red circle) -> disabled (left half of a filled red circle) -> removed (no breakpoint)
The shell and kernel only see the enabled breakpoints, but not the disabled/half breakpoints. And the menu "Shell -> Clear all {n} breakpoints" only counts and clears the enabled breakpoints.

The second part adds navigation via shortcuts (and menu entries). New default keys Alt+Up/Down will move the cursor in the editor along the arrow direction to the nearest row that has a full or half breakpoint marker. In the third commit I made the jumping behave like when pressing PageUp or PageDown keys: the horizontal position of the cursor will be kept instead of moving to the beginning of the row when jumping vertically.


bdieterm:
> Have you seen similar behavior (using breakpoints) in other IDE's that you took inspiration from?

Actually, no. In Notepad++ there are so-called "bookmarks" that are filled blue circles on the right of the line number. They can be set via Ctrl+F2 and navigated via F2 and Shift+F2, but there are no breakpoints.

I was also considering implementing proper bookmarks, but there are some things that shooed me away from that idea:
- bookmarks are expected to be persistent
Users would report a bug if the bookmarks vanish after closing a file or the whole application. Storing these extra data and synchronizing the lines (e.g. after running `python3 -m black somefile.py`) would add much complexity to the code base, making it prone to new bugs.
- bookmarks would require an extra shortcut and menu entry for toggling, extra shortcuts for navigating (unless refraining from navigating breakpoints), and extra style settings
- bookmarks would need either a second column for blue circles (wastes space) or require combined/overlayed symbols when bookmark and breakpoint are present at the same text line, and there is another overlay to consider: the debug marker (green ellipse). In case of using only one column, a context menu would be required when clicking next to the line number, for choosing between breakpoint and bookmark editing.

I just had a look at the KDevelop IDE. In the vertical bar next to the line numbers there are stars for bookmarks and red hexagons for breakpoints. The interesting thing is, that keyboard navigation via Alt+PageUp/PageDown treats all types the same: breakpoints, disabled breakpoints and bookmarks. This made me thinking of also using Alt+PageUp/PageDown instead of Alt+Up/Down -- I will add another commit for that.

bdieterm:
Just changed the shortcuts to the same used in KDevelop. And also updated the PR title to better reflect the changed/features introduced by this PR. Breakpoints that can be disabled are a common feature among various IDEs. Also breakpoint navigation via menu or shortcuts will then be the same as in KDevelop.

@almarklein: Are there any improvement suggestions or reasons for not merging from your part?

almarklein:
> or reasons for not merging from your part?

Just a lack of time, and my head spinning from other work 😄 

Ok, you bring forth compelling arguments. Let's do this 👍

almarklein:
Let's also do a release soon, we've had a lot of nice contributions recently.

PR #956, opened 2024-02-11
fix parsing/detection of "FIXME" in comments

bdieterm:
The "Source structure" tool listed ToDo-Type entries "todo" and "2do" but not "fixme".
This PR adds the "fixme" detection to the core codeparser so that it behaves the same as the codeparser for syntax highlighting in the editor.


almarklein:
Nice!

PR #957, opened 2024-02-11
increase search speed in File Browser tool

bdieterm:
This change will increase the speed of searching text via the File Browser tool by a factor 5 to 10 -- tested on Linux when searching in the Pyzo code base.
This PR sets `QUEUE_DELAY` to 0.1 ms (instead of 10 ms). The `IDLE_DELAY` value remains at 10 ms, so there should be no negative impact on CPU load or battery energy usage while not searching.


almarklein:
Oh wow 🚀

PR #958, opened 2024-02-12
fix shortcut strings

bdieterm:
I was wondering why some shortcuts like "Alt+PageUp" were not working in Pyzo.
It turned out that the string representation of the shortcut was wrong, at least for Linux and MS Windows with both Qt5 and Qt6 wrappers. For this example "PgUp" instead of "PageUp" fixes the problem.

@almarklein
Could you please test this on a Mac?

Actually there would be another solution to fix this:
```
ks = QtGui.QKeySequence(event.key())
print(ks.toString(ks.PortableText))
print(ks.toString(ks.NativeText))
# see https://doc.qt.io/qt-6/qkeysequence.html#SequenceFormat-enum
```
But I don't know how the portable and native texts would look on a Mac. Maybe this could replace the current dictionary `keymap`.


almarklein:
Intererestingly, Delete worked fine as it was, but indeed pageup (fn+up) and pagedown did not work, but they do with these changes. I also checked that `del` and a few other kept working with the shorter version.

Issue #946, opened 2024-01-01
cant' open a .py file with Pyzo and Pyzo crashed if config.ssdf is removed 


operade:
Hi I have just installed Pyzo with Homebrew on a Mac on which I did a clean install of macOS Sonoma 14.2.1 and two things are stranged

first : when I click on file say example.py a pyzo window appears with nothing except an empty shell. In can use the "open file" option in the menu to do it, but I didn't have to do that before.

second : I tried to remove the config file "config.ssdf" which is in the hidden directory .pyzo : bad idea because pyzo crashed although I was expected that the config file will be rebuilt. If I uninstall with homebrew and then reinstall Pyzo (without putting the config file) it is the same. I join a report file.
[report.txt](https://github.com/pyzo/pyzo/files/13805451/report.txt)

What did I did wrong ? Any help would be appreciated. Thanks in advance and Happy New Year.

David

almarklein:
When I try to open a `.py` file on my Mac by double-clicking, Pyzo opens but does not open the file. So it looks like something is broken there 🤔  - I did change something related recently: #923

When the "config.ssdf" is deleted, Pyzo should indeed revert to its defaults and overwrite the config file when it exists. The crash report suggests that it segfaults, which is pretty bad and is usually related to Qt.

Have you tried the binary from Github, or running Pyzo from source?

operade:
Hi Installing Pyzo with the .dmg file is the same for both issues; I haven't tried the Github binary. However, at the beginning I wanted to change the config.ssdf file because I was unable to change the path where Python is located on my Mac in the Pyzo preferences pane : if I did so, when Pyzo restarted the change was not made. I was able to resolve the issue by changing the owner of the file from "root" to "myself". Best regards

almarklein:
So the crash is fixed, and the remaining issue is that double-clicking a file in the OS file explorer does not open the file in Pyzo, am I right?

operade:
Yes absolutely : sorry for my late reply

jowe19:
Hi. 
I've got the same problem ; if I try to open a `py` file on my Mac by double-clicking, it launches Pyzo but without opening the file.

bdieterm:
@jowe19
I don't have a Mac, but you could try the following to hopefully get more information about the problem:
Start Pyzo normally.
Open the "Logger" tool (via "Tools -> Logger" in Pyzo's menu).
Double click a Python file to open it.
Normally the Logger will show some text like:
Request: /path/to/file.py
[...]

PR #959, opened 2024-02-16
fix file opening via QFileOpenEvent on Macs

bdieterm:
As reported via https://github.com/pyzo/pyzo/issues/946#issuecomment-1947920124, files cannot be opened anymore via file association on Mac computers.
```
# in the callback for QFileOpenEvent:
fname = '/path/to/script/tobeopened.py'
sys.argv = ['/path/to/pyzo']
sys.argv[-1:] = [fname]  # <-- bug is here
res = commandline.handle_cmd_args()

# in function handle_cmd_args:
args = sys.argv[1:]  # <-- args is then an empty list
```

This seems to be a bug introduced by #923.
In #923, two files were changed:
I fully know about the problem and its solution that lead to changes in file `pyzo/pyzokernel/start.py`.
But I have no idea, what the reason was to also modify file `pyzo/_start.py` in that PR.

### DIFF
--- a/pyzo/_start.py
+++ b/pyzo/_start.py
@@ -69,7 +69,7 @@ class MyApp(QtWidgets.QApplication):
         if isinstance(event, QtGui.QFileOpenEvent):
             fname = str(event.file())
             if fname and fname != "pyzo":
-                sys.argv[-1:] = [fname]
+                sys.argv[1:] = [fname]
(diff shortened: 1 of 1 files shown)


almarklein:
I guess in #923 I figured to prevent args from getting lost, but apparently that had unforeseen consequences. Thanks for tracing this down!

PR #962, opened 2024-02-23
fix Python syntax parser and quotes/brackets autocompletion

bdieterm:
Actually, I only wanted to fix the following bug:
Syntax highlighting for a normal string literal that has a line-continuation backslash does not work, e.g.:
```python3
s = 'abc\
def'
```
While fixing this and testing, I encountered several bugs that are related (and were already present before this PR).

The first commit adds some missing bult-in identifiers for newer Python versions.
The second commit fixes some comments for the parser regular expressions that probably got mixed up when reformatting the code.
The third commit adds the line-continuation with backslash for normal strings, as described at the beginning.

When testing auto-completion with the line-continued string I saw that, when entering `[`
```python3
s = 'abc\
def  [   \
'
```
autocompletion added the `]` which should not happen inside string literals.
So I fixed the bug in method `_get_token_at_cursor` via the fourth commit.

With the fifth commit I fixed the parser so that it properly recognizes string literals with prefixes with valid prefix combinations, e.g. `rf"abc"`.
Then I did some testing again, also with autocompletion enabled. I normally do not have quotes and brackets autocompletion enabled, so I was surprised that quotes autocompletion never has worked for prefixed string literals. For example `s = r'` will not close the single quote automatically because the "r" was parsed as an identifier (because the single quote was too new). I fixed this, together with some other adjustements, via the sixth commit.

With these changes, the syntax highlighter now formats the prefix for string literals in the same color/format as the remaining string literal. In the official Python docs, and in some code editors, the prefix is also included in syntax highlighting. Autocompletion will now also complete brackets directly after a string literal, e.g. `"xyz"[` will turn into `"xyz"[]`.

### DIFF
--- a/pyzo/codeeditor/extensions/behaviour.py
+++ b/pyzo/codeeditor/extensions/behaviour.py
@@ -22,7 +22,7 @@ from ..parsers.tokens import (
     StringToken,
     UnterminatedStringToken,
 )
-from ..parsers.python_parser import MultilineStringToken
+from ..parsers.python_parser import MultilineStringToken, stringLiteralPrefixes
 
 
 class MoveLinesUpDown(object):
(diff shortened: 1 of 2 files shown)


almarklein:
wow ... 🙏

PR #965, opened 2024-02-27
fix for fix initial horizontal scrollbar positions for editors

bdieterm:
Sorry for the inconvenience. I just discovered that this did not work on the MS Windows version. Fixed that.


almarklein:
😄 thanks for the quick fix!

PR #969, opened 2024-03-13
fix detection of already opened files

bdieterm:
The current detection of already opened files solely relies on case sensitivity.

With the new implementation, symbolic links and network addresses (`r"\\localhost\c$\temp\abc.py" and r"C:\temp\abc.py"`) will also be recognized as the same file.


pyzo/core/baseTextCtrl.py
```diff
@@ -25,41 +25,12 @@ def normalizePath(path):
     """Normalize the path given.
     All slashes will be made the same (and doubles removed)
```
almarklein, line 26: This function is used in a few other places, but I think its ok to reduce it like you did here.

PR #973, opened 2024-03-24
add "Execute line as statement" to the run menu

bdieterm:
Why another execution command that could confuse users? 
Because this fills a gap for interactively working with code, when adding or modifying the code in the editor (instead of the shell).
I often find myself in the situation that I want to evaluate the current line in the editor and see the result printed in the shell.
Currently this was only possible by copying the line from the editor to the shell via clipboard, or by selecting the whole line and pressing F9 (or Alt+Return). When F9 is pressed without a selection, the current line in the editor will be executed as well, but the shell will only print `>>> (executing line xy of "my_script.py")`.
The new command has the F10 keyboard shortcut assigned.
I also added the description to the Pyzo wizard. The explanation is simple: Copy the whole current line from the editor to the shell and execute it, the same way it would happen when doing this manually via the clipboard.


almarklein:
I definitely can see it's use.

Issue #841, opened 2023-02-05
Dark mode and problems about "match-case"


JimmyXiaoShort:
I want Pyzo to have dark mode, just like spyder.
At the same time, I hope that Pyzo can correctly handle the highlighting of "match-case" statements.
This statement has been added in Python3.10 Beta, but Pyzo does not support highlighting it as a keyword or in the complete List.

Pavel-Sayekat:
![Screenshot at 2023-02-25 12-32-24](https://user-images.githubusercontent.com/17138178/221342493-0cf4e565-d903-4ef9-b32e-ddda6df64acf.png)

Dark mode like this?

JimmyXiaoShort:
Yes.
Jimmy Xiao Short从Windows邮件发送

发件人: Pavel ***@***.***>
发送时间: 2023年2月25日 14:33
收件人: ***@***.***>
抄送: Jimmy Xiao ***@***.***>; ***@***.***>
主题: Re: [pyzo/pyzo] Dark mode and problems about "match-case" (Issue #841)


[Screenshot at 2023-02-25 12-32-24]<https://user-images.githubusercontent.com/17138178/221342493-0cf4e565-d903-4ef9-b32e-ddda6df64acf.png>

Dark mode like this?

―

JimmyXiaoShort:
Thank you for your attention!

Best wishes to you and your pyzo.

发件人: Pavel Sayekat ***@***.***>
发送时间: 2023年2月27日 17:07
收件人: pyzo/pyzo ***@***.***>
抄送: Jimmy Xiao Short ***@***.***>; Author ***@***.***>
主题: Re: [pyzo/pyzo] Dark mode and problems about "match-case" (Issue #841)


Goto "settings" --> "Edit syntax styles" and select the syntax color mode you like, and for windows appearance, goto "View" --> "Qt theme" and select the theme you like, I used kvantum tool to select a dark theme and then select kvantum from "Qt theme" list.

bdieterm:
>I want Pyzo to have dark mode, just like spyder.
At the same time, I hope that Pyzo can correctly handle the highlighting of "match-case" statements.
This statement has been added in Python3.10 Beta, but Pyzo does not support highlighting it as a keyword or in the complete List.

Dark mode was improved in #917.
Match case highlighting was implemented in #980.
Autocompletion was implemented in #1132.

PR #980, opened 2024-03-30
add support for soft keywords "match" and "case"

bdieterm:
This will add proper syntax highlighting for the keywords "match" and "case" but only if they are used in such a way.
See the doc string of method `_promoteMatchCaseSoftKeywords(tokens)` for details and examples.

This will also close issue #841.


almarklein:
Nice!  IIUC the complexity of conditionally promoting keywords is that they you also have code like `foo.match(..)`, right?

bdieterm:
Exactly. Very common calls such as `re.match(...)` would not work anymore if `match` would have been changed to a keyword such as `else` instead of a "soft keyword". There is section "Hard vs. Soft Keyword" in [PEP-635](https://peps.python.org/pep-0635/) which goes into more detail.

Issue #989, opened 2024-05-09
Python shell with asyncio event loop crashes when pressing the "Interrupt" icon


bdieterm:
Since the switch to the new experimental asyncio event loop (see #789), the shell crashes when pressing the "Interrupt" icon.

The crash happened on all tested operating systems and Python interpreters (kernel):
MS Windows, Linux, Python 3.10 to 3.13

Steps to reproduce:
* edit the shell configuration to have the "Asyncio" event loop actived.
* start a new shell
* press the "Interrupt" icon (or press "Shell -> Interrupt" in Pyzo's menu)

This problem is quite bad because "asyncio" is automatically selected as the event loop for a freshly installed Python interpreter, and this leaves a bad first impression for new users.

@almarklein:
You are much more familiar with that asyncio stuff than me. Could you please have a look at this. Meanwhile I will create a PR that will change the asyncio event loop back to the old implementation (with additional fixes for Python >= v3.12).

bdieterm:
Fixed by @almarklein.

PR #990, opened 2024-05-09
switch back to old asyncio event loop implementation

bdieterm:
This commit will use the old asyncio event loop implementation again instead of the new experimental one that causes a crash when interrupting exection -- see issue #989. I also updated the code for newer Python versions.


bdieterm:
obsolete because of better solution #991

PR #991, opened 2024-05-10
Catch keyboardinterrupt in asyncio loop

almarklein:
Alt solution to #989. It catches the interrupt exception and restarts the loop.


bdieterm:
Perfect! Thanks a lot.

PR #995, opened 2024-05-13
Fix foreground colour not resetting in the shell when using the ANSI foreground reset code

### LINKED ISSUE #996: Foreground colour does not reset in the shell when using the ANSI foreground reset code
When using ANSI colour codes in the shell, putting a reset code doesn’t actually reset the colour.

## How to reproduce?

Requires `colorama` but can be performed manually
```python
>>> from colorama import Fore
>>> print(Fore.RED + "hello" + Fore.RESET, "world")
```

Expected result: “hello world” with “hello” in red.

Actual result: “hello world” entirely in red.

I propose a simple fix in #995

jd-develop:
The foreground colour was not resetting in the shell when using the ANSI foreground reset code. Fixes #996

### DIFF
--- a/pyzo/core/shell.py
+++ b/pyzo/core/shell.py
@@ -907,6 +907,8 @@ class BaseShell(BaseTextCtrl):
                 elif 30 <= param <= 37:  # Set foreground color
                     clr = CLRS[param - 30]
                     format.setForeground(QtGui.QColor(clr))
+                elif param == 39:  # Reset the foreground color
+                    format.clearForeground()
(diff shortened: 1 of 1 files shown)


bdieterm:
Thanks for the contribution, @jd-develop!

In case anyone wants to test this without installing package "colorama", then try this:
`print('abcd' '\x1b[0;1;4;31m' 'hello' '\x1b[39m' 'world' '\x1b[0m' 'efgh')`

almarklein:
Thanks!

PR #1000, opened 2024-05-26
add pause feature for interrupting and resuming code execution

bdieterm:
Currently, executed code can be interrupted by the user via "Shell -> Interrupt (Ctrl+I)" or by setting breakpoints. The former way only allows post-mortem debugging but not continuing execution after the interruption. The latter way (breakpoints) does allow resuming execution, but before running the code, at least one breakpoint has to be present, and this will activate tracing which results in a slow-down of the execution by a factor of up to 50 (depending on the code). The slower code execution also remains after continuing execution with all breakpoint removed.

This PR will add a new feature for code execution in Pyzo shells to pause code execution.
To use this, the user has to press "Shell -> Pause" in Pyzo's menu or click the new icon in the shell's toolbar. This will interrupt the code execution by sending signal "SIGFPE" to the process that runs the Python kernel. Once interrupted, the debug mode will be activated the same way as for a triggered breakpoint. Resuming the code execution is also the same as for breakpoints. Pausing and resuming the same code execution multiple times is also no problem.
The interruption can occur in any code line, wherever the executed code is located at the moment the user hits the Pause button. To have better control of where to stop, the pause feature can be used together with breakpoints. For that I have modified the behavior of breakpoints slightly.

Modified breakpoint behavior (not related to the new pause feature):
This is still the same:
When executing code and at least one breakpoint was set before, the debugger will activate tracing and stop once a breakpoint is reached. Breakpoints can be added or removed during interruptions.
This is modified:
When, during an interruption, all breakpoints are removed by the user, tracing will be deactivated to allow full speed execution.

When the interpreter runs in full speed mode, i.e. without tracing, breakpoints can be added even after the execution was started. Simply set the breakpoint(s) and press the new pause button (or the other way round). Continue the execution and the breakpoints will be updated, and tracing will be enabled if there is at least one breakpoint resp. disabled if there are no breakpoints.

Why did I choose SIGFPE (Floating-Point Exception Signal)?
This seems to be the only signal available on all operating systems and to be a signal that is actually never used (at least for Pyzo with Python kernels). Why do I know that this signal is actually never used? I tried this on my Linux system: Sending SIGFPE to the Python kernel's process will immediately exit that process (tested with the current Pyzo release).
Because of sending this signal, this pause feature will require Python >= v3.10 on a MS Windows OS. But for Linux, this even works with Python v2.7.

I have tested this a bit on Linux and MS Windows with different Python interpreters (also PyPy and with IPython) and have not ecountered any issues so far.
@almarklein: Could you please do a quick check if there are any issues on a macOS computer (post-mortem, breakpoints, pause)?


almarklein:
This is really awesome!

I tested this out and it seems to work fine, and I have not observed any regressions.

However, I did notice that I cannot pause when in callbacks in interactive mode. To illustrate, I created this minimal example, for use with GUI integration set to "asyncio":

```py
## This works fine. Can interrupt, which brings me inside foo()

def foo():
    i = 0
    while True:
        i += 1

foo()

## Cannot interrupt to get into bar()

import asyncio

def bar():
    x = []
    for i in range(100):
        x.append(i)
    # print("done")
    loop.call_soon(bar)

loop = asyncio.get_running_loop()
loop.call_soon(bar)
```

bdieterm:
> I did notice that I cannot pause when in callbacks in interactive mode.

Now, with #1003, pausing the callback in the interactive mode is also possible.

almarklein:
Amazing! This new feature adds a whole new level of interactivity!

PR #1009, opened 2024-06-09
add jump feature to debugger

bdieterm:
Up to now, continuing execution after interruption (via breakpoints or the new "pause" feature) was only possible at the line where it has been stopped.
With this new "jump" feature, the line where execution should continue, can be changed while the debugger is active. Similar to the other debug commands, there are two ways to invoke this:

1. via the "Shell -> Debug jump" in Pyzo's menu bar, or the new button in the shell's toolbar
The line number, where execution should continue, will be the line of the text cursor in the currently selected code editor.
2. or via the magic debug command: just enter `db jump 23` where `23` is an example value for the line number.

In the debug mode, the stack frame has to be set to the lowest frame (the one with the highest number) before using the feature. Right after the interruption, the lowest frame is selected anyways.
It is not possible to do special jumps like jumping inside the middle of a for loop, as mentioned [in the official Python docs](https://docs.python.org/3/library/pdb.html#pdbcommand-jump).

Jumping is possible in both directions: You can jump to one of the previous lines to execute these lines again, or you can skip the execution of some lines.

Here is a simple example for anyone who wants to test this. Just set a breakpoint somewhere in the function and run the whole code.
```python3
ll = []
def myfunc():
    ll.append('a')
    ll.append('b')
    ll.append('c')
    ll.append('d')
    ll.append('e')

myfunc()

print(ll)
```
I have tested this with several Python interpreters (e.g. 3.11, 3.12, 3.8, 2.7, PyPy) on Linux and also with 3.12 on MS Windows.
It does not matter if the debug mode was started via a normal breakpoint, the `breakpoint()` function or the new "pause" feature.


almarklein:
🤯

PR #1011, opened 2024-06-10
modernize Python2-style code to Python3

bdieterm:
The commits in this PR will modernize the code base of Pyzo's IDE components, but without the pyzokernel and yoton modules. Running a Python 2.7 kernel is still possible after this PR.

This PR also corrects some minor bugs, for example the following two lines that were missing the values for the format strings:
`print("Unable to delete Python version %s: does not exist.")`
`self._problem = "%s is not a valid file."`


almarklein:
Nice :)

PR #1018, opened 2024-06-12
fix jumping for code execution with line offset

bdieterm:
With this fix, jumping in code that was executed as a cell or as a selection will also work correctly.


almarklein:
You're on a roll! 🚀

PR #1030, opened 2024-06-23
improve auto-completion and call-tips, and add key-completion

bdieterm:
This PR adds some handy auto-completion features to Pyzo:

- better working call-tips for complex expressions like `abc = (myfunc(x + ((y + 2) + otherfunc(1, 2) - `
- key auto-completion (for dictionaries, lists and similar objects)
- auto-completion and calltips using a cached evaluated object, using the new shortcut Ctrl+Space

Examples for cached auto-completion:

Type `locals().` (with the **dot** at the end) and press Ctrl+Space. This will evaluate `locals()` and assign the result to `__pyzo__autocomp` in the local scope. Then auto-completion continues and displays the list widget with the **attributes**.

Type `locals()[` (with the **open bracket** at the end) and press Ctrl+Space. This will evaluate `locals()` and assign the result to `__pyzo__autocomp` in the local scope. Then auto-completion continues and displays the list widget with the **keys**.

Type `locals().get(` (with the **open parenthesis** at the end) and press Ctrl+Space. This will evaluate `locals()` and assign the result to **`__pyzo__calltip`** in the local scope. Then auto-completion continues and displays the **calltip** for `dict.get(...)`.

The new Ctrl+Spacehot key can also be pressed while having already entered some characters of the attribute, key or arguments, in case someone cancelled the autocompletion list widget resp. calltip and wants to show it again

I chose this approach with manually pressing the new hotkey instead of automatically evaluating the expression because the latter could unintendedly and silently modify the data, for example with the `pop()` in `mylist = [1, 'b', ()]; mylist.pop().count(`.
The old auto-completer for attribute-look-up also evaluated expressions during auto-completion, but only for simpler expressions and without caching -- properties and methods like `__getattribute__` were called multiple times in the background, and still are, but only for simple expressions.
Key-auto-completion will, like attribute-auto-completion and calltips, also work for simple expressions without pressing a hot-key.

### DIFF
--- a/pyzo/core/baseTextCtrl.py
+++ b/pyzo/core/baseTextCtrl.py
@@ -30,75 +30,281 @@ def normalizePath(path):
     return path
 
 
-def parseLine_autocomplete(tokens):
-    """Given a list of tokens (from start to cursor position)
-    returns a tuple (name, needle).
-    autocomp_parse("eat = banan") -> "", "banan"
-      ...("eat = food.fruit.ban") -> "food.fruit", "ban"
-    When no match found, both elements are an empty string.
+def isFormatStringToken(token):
+    """returns True if a StringToken is a format string literal, e.g. f'{1 + 2}' """
+    if not isinstance(token, Tokens.StringToken):
+        return False
+    s = str(token)
+    return "f" in s[:s.index(s[-1])].lower()
+
+
+def isSimpleExpression(tokens):
+    """returns True if the list of tokens is a simple expression
+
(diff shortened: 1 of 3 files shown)


almarklein:
Very nice!

> I chose this approach with manually pressing the new hotkey instead of automatically evaluating the expression because the latter could unintendedly and silently modify the data

👍  Generally speaking, for attribute and key access one can generally assume the result is fast and without side-effects. For function calls we cannot know what will happen. We don't know how long it will take and what side-effects it will have.

PR #1040, opened 2024-06-29
Fix yoton

bdieterm:
This PR includes minor fixes and some cleanup for the yoton module.


almarklein:
This breaks compatibility with Python 2.4 😅 

I started programming in Python when it was in version 2.4, and tried to keep Pyzo compatible with that for a long time. Time to let go ;)

PR #1045, opened 2024-06-30
remove PyzoAssistant

bdieterm:
This PR will remove the "PyzoAssistant" that was started via "Help -> Local documentation" in Pyzo's menu.

As far as I have seen, the scripts by Windel Bouwman to generate the compiled help files are quite outdated (e.g. referring to SciPy v0.14 on sourceforge). The size of that extra help window is also quite small and does not properly scale. I don't see a value in that tool -- The Interactive help tool and a normal web browser are pretty good already.

@almarklein: Are there any arguments to keep that PyzoAssistant?
And by the way: How about a new release in maybe one or two weeks?


almarklein:
Yes, I think this can go.

> How about a new release in maybe one or two weeks?

Yeah, so many awesome improvements! Really great work!

almarklein:
@bdieterm Any further plans for pr's before we do a release? If not I'll probably do it tomorrow.

bdieterm:
Doing the relase tomorrow is fine.
Are the macOS specific problems #927 and #943 still relevant or are they solved with the newest PySide6 version? The macOS build is currently pinned to PySide6==6.6.0 and the newest PySide6 release is 6.7.2.

almarklein:
Good question. It appears that the problems mentioned in these prs are not present in the latest pyside6. Following up in #1049. Also looking into building for MacOS aarch64 (Apple silicon), since the default MacOS runners are M1 nowadays...

PR #1058, opened 2024-08-04
add expression viewer tool

bdieterm:
This PR adds a new tool "Expression viewer", updates the "Workspace" tool and adds the new method `evalMultiple`  in the introspector for evaluating multiple expressions used by the Expression viewer tool.

The "Expression viewer" is a kind of variable watch window that can be used for any expressions (not just variables). This also allows for custom formatting etc. by simply adding the python code for that in the expression. The results are normally updated automatically, the same as in the "Workspace" tool. For manual updates (e.g. when code is executed in the event loop), a manual refresh can be performed.

The GUI of the new tool consists of a small tool bar with buttons and a list resp. text widget.
The button with the pencil is a toggle button to enable the "edit mode".
In the "edit mode", a text box will be displayed instead of the 2-column table, and the expressions can be edited, each on its own line. Inserting/removing a "#" prefix at the start of the line will disable/enable the expression -- this will be reflected in the state of the checkbox in the list view.
Expressions can also be modified in the list view, but for manually adding/removing expressions, the "edit mode" is recommended.
The expression viewer tool does not save any configuration data. To quickly save and restore the expressions, switch to the edit mode and copy/paste the contents to/from a text file.

The values of the expressions will be displayed as the truncated str representation (not repr).
But the context menu of the list view has a "Toggle repr" action which edits the expression: A repr command will be added or removed. This is useful for multiline-strings and special datatypes (e.g. datetime and numpy modules).

The "Show in edit mode" action of the context menu is intended for quickly deleting or editing the selected entry.

Another way to add an expression is to select it in the editor and press the "add selected expression ..." button in the Expression viewer tool.

The results can be copied by selecting the contents of the cell in the "Result" column. For obtaining the full string, the "Evaluate in shell" action (context menu) can be used.

The Workspace tool got a new command "Add to Expression viewer tool" in the context menu that adds the selected variable in the list view to the Expression viewer tool. This can also be done by pressing the "v" key. And there is also a new "add all visible entries to the Expression viewer tool" button.

The refresh buttons of the Expression viewer tool will also trigger the refresh of the Workspace tool and vice versa.

### DIFF
--- a/pyzo/pyzokernel/introspection.py
+++ b/pyzo/pyzokernel/introspection.py
@@ -459,6 +459,33 @@ class PyzoIntrospector(yoton.RepChannel):
         # Done
         return text
 
+    def evalMultiple(self, sequenceOfExpressions):
+        """evaluate multiple expressions
+
+        The result will be converted to a string and truncated if too long.
+        In case of an error, the result will be the exception as a string.
+
+        return value:
+            [(request1, success1, result1), (request2, success2, result2), ...]
+        """
+        NS = self._getNameSpace()
+        maxChars = 150
+
+        returnValue = []
+        for expr in sequenceOfExpressions:
+            try:
+                result = str(eval(expr, None, NS))[:maxChars + 1]
+                if len(result) > maxChars:
+                    result = result[:maxChars - 1] + "…"
+                success = True
+            except Exception as e:
+                result = str(e)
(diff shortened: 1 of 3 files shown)


almarklein:
Oh nice! Will check this  out when I'm back from my holidays.

Issue #470, opened 2017-02-22
Keyboard shortcuts not working on Linux with Pyqt5


maria8899:
Hi,
I cannot copy and paste code using the keyboard within the editor or from editor to shell.

Version info
Pyzo version: 4.3.1 (source)
Platform: linux
Python version: 3.5.2+
Qt version: 5.8.0
PyQt5 version: 5.8

Pyzo directories
Pyzo source directory: /usr/local/lib/python3.5/dist-packages/pyzo
Pyzo userdata directory: /home/maria/.pyzo

Any idea why?

almarklein:
This has recently been reported: https://groups.google.com/forum/#!topic/pyzo/HFoBWJ97zsg

It seems to be an issue related to a new version in Qt or PyQt. Maybe a certain part of the API changed, breaking our copy-paste functionality.

almarklein:
Spun up a Linux VM.

With Python provided by miniconda, I don't see the problem. Version info:
Pyzo version: 4.3.1 (source)
Platform: linux
Python version: 3.5.3
Qt version: 5.6.2
PyQt5 version: 5.6

With the system Python and Pyqt5, I can reproduce the problem. Versiom info:
Pyzo version: 4.3.1 (source)
Platform: linux
Python version: 3.4.3
Qt version: 5.2.1
PyQt5 version: 5.2.1

So it seems to be related to PyQt5, though oddly enough the problem does not seem to happen with miniconda/anaconda. It does not seem to be related to a specific version of qt5 though; Maria's qt version and my qt version differ quite a bit, and my miniconda version is in between them.

As a workaround, one can switch to using Pyqt4, by `apt-get install python3-pyqt4`, and `export QT_API=pyqt4`. Or one can use the miniconda/anaconda interpreter to launch Pyzo.

almarklein:
Actually, this is not specific to copy-paste, but to *all* shortcuts.

almarklein:
Fixed by https://github.com/pyzo/pyzo/commit/0064a5c709e4c5f6f67832128f022469c637a4c2

This Qt problem has been reported before (http://stackoverflow.com/questions/23916623/qt5-doesnt-recognised-shortcuts-unless-actions-are-added-to-a-toolbar). There its described as a bug, but since its still there in qt 5.8, its more like a backward incompatible change.

People suffering from this bug can add that line to their local version of Pyzo.

maria8899:
Great, adding the line has fixed this issue. Thank you!

PR #1061, opened 2024-08-07
remove old workaround for shortcuts now causing issues

### LINKED ISSUE #1059: Cannot create working shortcut for selecting any tool widget
I want to select active tools via shortcuts and added them via "Shortcut mappings/tools" menu.

I originally tried setting them to various F-keys, but also tried "Ctrl+Fx" or even a simple "Ctrl+H". 
I deleted all other shortcuts using F-keys above F3 (so there are no duplicates).

No matter which shortcut I try for any tool selecting action, the action is NOT getting fired, but an error appears in the pyzo startup console on every shortcut press:
```QAction::event: Ambiguous shortcut overload: [...]```
With `[...]` being the name of some shortcut pressed for tool selection (e.g. 'F5', 'Ctrl+F5' or 'Ctrl+H' etc).

I'm not sure how to interpret this error, but it seems to hint to a bug while handling shortcuts for these kinds of actions.
I only found this so far:
https://forum.qt.io/topic/99653/ambiguous-shortcut-overload

**Quote from the answer:**
```3. The first things that comes to my mind would to check what you do with all your actions. For example, are they set on several widgets ? Are they re-used ?```

**Maybe it has to do with the fact, that these actions are present in 2 different menus (menu bar, context menu of tools)!?**
grafik
grafik

Has somebody tried and succeeded in using shortcuts for tool selecting actions on any platform or can reproduce this issue?

### Environment:
Windows-x64
Python 3.9.17-x64
PySide2 5.15.2.1
pyzo 4.16.0

e-d-n-a:
Old issue #470 seems to be fixed in Qt 5.15 LTS and Qt 6. Old workaround causes error "QAction::event: Ambiguous shortcut overload" now in latest Qt versions, at least when creating shortcuts for actions like "tool selection".

Fixes #1059.

### DIFF
--- a/pyzo/core/menu.py
+++ b/pyzo/core/menu.py
@@ -236,9 +236,6 @@ class KeyMapper(QtCore.QObject):
             # Set shortcut so Qt can do its magic
             shortcuts = pyzo.config.shortcuts2[action.menuPath]
             action.setShortcuts(shortcuts.split(","))
-            pyzo.main.addAction(
-                action
-            )  # issue #470, http://stackoverflow.com/questions/23916623
             # Also store shortcut text (used in display of tooltip
             shortcuts = shortcuts.replace(",", ", ").replace("  ", " ")
             action._shortcutsText = shortcuts.rstrip(", ")


bdieterm:
Perfect. Thanks!

Issue #1088, opened 2024-10-11
PySide6 6.8.0 causes a segmentation fault


bdieterm:
Currently, Pyzo builds with PySide6 6.8.0 (which was released two days ago) fail.
I was able to narrow down the problem to a single line:
`import PySide6; from PySide6.QtCore import *`

I already reported this problem in the Qt bug tracker.

bdieterm:
Solved via workaround #1091.

PR #1091, opened 2024-10-11
rework the Qt wrapper to work around a bug in PySide6 6.8.0

bdieterm:
As a workaround for issue #1088, I reworked Pyzo's Qt wrapper to not use star imports (`from ... import *`) anymore.


almarklein:
Nice work!

PR #1100, opened 2024-11-15
Bump version

almarklein:


almarklein:
Had to agree to the latest Apple developer agreement, but binaries are building!

PR #1105, opened 2024-12-09
Enable pm for logging again

almarklein:
Follow-up on #1103

The ability for setting up PM debugging when a ``logger.error()`` is quite generally useful IMO. Basically any case where there is some sort of event-loop that catches and logs exceptions instead of letting them fall through. In any case, I make use of it a lot :)

I took some time to look into what the least invasive approach could be, and come to the conclusion that adding a handler will always hurt one use-case or another. So I applied a somewhat dirtier monkey-patch, but that otherwise leaves the logging system alone.

### DIFF
--- a/pyzo/pyzokernel/interpreter.py
+++ b/pyzo/pyzokernel/interpreter.py
@@ -404,8 +404,12 @@ class PyzoInterpreter:
                 sys.last_type, sys.last_value, sys.last_traceback = record.exc_info
             return callHandlers_ori(self, record)
 
-        callHandlers_ori = logging.Logger.callHandlers
-        logging.Logger.callHandlers = callHandlers_patched
+        try:
+            callHandlers_ori = logging.Logger.callHandlers
+            logging.Logger.callHandlers = callHandlers_patched
+        except Exception:
+            pass
+
 
     def _handle_sigfpe(self, sig, frame):
         self.debugger.set_trace(frame)
--- a/pyzo/pyzokernel/interpreter.py
+++ b/pyzo/pyzokernel/interpreter.py
@@ -392,7 +392,7 @@ class PyzoInterpreter:
         #
         # Adding a handler to the root logger (as we did earlier) seems like
         # an elegant approach, but it gets complicated as to whether Pyzo should
-        # take responsibility for pinting the message. The fact that a handler
+        # take responsibility for printing the message. The fact that a handler
         # is present, will stop the logging module from printing the message,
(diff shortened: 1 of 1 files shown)


pyzo/pyzokernel/interpreter.py
```diff
@@ -387,49 +384,33 @@ def parse_version_crudely(version_string):
...
-        addLoggerPostMortemHandler()
+        # Setup post-mortem debugging via appropriately logged exceptions.
+        #
+        # The Logger.callHandlers look like a good place to Monkey-patch, as
+        # it's certain that the record is going to be handled at this point,
+        # and it does not interfere with the rest of the logging mechanics.
+        #
+        # Adding a handler to the root logger (as we did earlier) seems like
+        # an elegant approach, but it gets complicated as to whether Pyzo should
+        # take responsibility for pinting the message. The fact that a handler
```
bdieterm, line 395: typo: `pinting` -> `printing`

pyzo/pyzokernel/interpreter.py
```diff
@@ -387,49 +384,33 @@ def parse_version_crudely(version_string):
...
-        self._original_logging_basicConfig = None
         try:
-            self._original_logging_basicConfig = logging.basicConfig
-            logging.basicConfig = basicConfigPatched
+            callHandlers_ori = logging.Logger.callHandlers
+            logging.Logger.callHandlers = callHandlers_patched
         except Exception:
             pass
 
+
```
bdieterm, line 413: `flake8` complains about that extra blank line

bdieterm:
This is a very clever solution to the problem, much better than my previous fix. I just tested it and it works great.
Sorry for breaking PM debugging for that event loop use cases. [EAFP](https://docs.python.org/3/glossary.html#term-EAFP) ;-)

almarklein:
@bdieterm what are your thoughts on an autoformatter? Manually fixing formatting errors feels archaic 😅. I've been switching to `ruff` laterly for some projects (to do both linting and formatting) and am *very* pleased with it.

bdieterm:
I just tried `ruff check` in the Pyzo source directory.
There are many false alarms with the default configuration, but it found a real minor bug (see #1107).
By having to fix wrong code style manually, `flake8` made me change my behavior, for example putting two spaces before a line comment. Therefore autoformatting will not have a big impact for me.
I am fine with switching to `ruff`. Let me know if I can support you with that (configuration file, updating code, CI workflow, ...).

PR #1108, opened 2024-12-13
fix traceback for Python < v3.10 and for temporary files

bdieterm:
PR #1104 broke exception formatting for older Python versions.
When starting a shell with Python < v3.10 and evaluating a variable that does not exist, instead of the `NameError` another error is displayed in the shell:
```
>>> a
An error occured, but then another one when trying to write the traceback: format_exception() takes at least 3 arguments (1 given)
```
And there was also another small bug: Line `for i in range(max(len(lines), 5)):` would iterate over at least 5 lines even if there were less, and then cause an IndexError. Probably that should have been a `min` instead of the `max`.

The third problem with PR #1104 was that it broke line number correction for temporary files:
The tracebacks then had
`File "<tmp 3>+4", line 6, in <module>`
instead of
`File "<tmp 3>", line 10, in <module>`.

With this PR, all these problems are fixed.

To reproduce the problem and check the fix, I used the following code, in both a temporary (unsaved) file and in a saved file:

```python3
# just some empty lines so that we have a line offset when executing cells


##


# execute this cell to cause a ZeroDivisionError

1 / 0

##

# execute this cell to raise multiple exceptions as a group
# ExceptionGroup was introduced in Python 3.11

e = ExceptionGroup('exception group example', [ValueError(1), TypeError(2)])

raise e

# this prints the traceback with the sub-exceptions
```

### DIFF
--- a/pyzo/pyzokernel/interpreter.py
+++ b/pyzo/pyzokernel/interpreter.py
@@ -1172,34 +1172,28 @@ class PyzoInterpreter:
         try:
             if useLastTraceback:
                 # Get traceback info from buffered
-                type = sys.last_type
+                etype = sys.last_type
                 value = sys.last_value
                 tb = sys.last_traceback
             else:
                 # Get exception information and remove first, since that's us
-                type, value, tb = sys.exc_info()
+                etype, value, tb = sys.exc_info()
                 tb = tb.tb_next
 
                 # Store for debugging, but only store if not in debug mode
                 if not self._dbFrames:
-                    sys.last_type = type
+                    sys.last_type = etype
                     sys.last_value = value
                     sys.last_traceback = tb
 
-            lines = traceback.format_exception(value)
-
-            # Remove exec() call in interpreter.py"
-            for i in range(max(len(lines), 5)):
-                if "pyzo" in lines[i] and "interpreter.py" in lines[i] and "exec(" in lines[i]:
-                    lines.pop(i)
-                    break
+            lines = traceback.format_exception(etype, value, tb)
 
             for line in lines:
                 line = self.correctfilenameandlineno_on_line(line)
                 self.write(line)
 
         except Exception:
-            type, value, tb = sys.exc_info()
+            etype, value, tb = sys.exc_info()
             tb = None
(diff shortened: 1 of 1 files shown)


pyzo/pyzokernel/interpreter.py
```diff
@@ -1172,34 +1172,28 @@ def showtraceback(self, useLastTraceback=False):
...
                     sys.last_value = value
                     sys.last_traceback = tb
 
-            lines = traceback.format_exception(value)
-
-            # Remove exec() call in interpreter.py"
-            for i in range(max(len(lines), 5)):
-                if "pyzo" in lines[i] and "interpreter.py" in lines[i] and "exec(" in lines[i]:
-                    lines.pop(i)
-                    break
```
almarklein, line 1195: This is completely removed, but that means the line showing Pyzo internals will be shown in the tracebacks, I think?

almarklein (commented):
Thanks!  I should have tested with an older Python, sorry about that!

pyzo/pyzokernel/interpreter.py
bdieterm: No. As in older Pyzo versions, Pyzo's stack frame is already removed a few lines before that:
https://github.com/pyzo/pyzo/blob/b46302d5f14d2771e09336847ed4b252432fe3df/pyzo/pyzokernel/interpreter.py#L1173-L1175

With this new fix, the modified traceback object `tb` is passed to `format_exception`:
`lines = traceback.format_exception(etype, value, tb)`
We need to pass it anyways, because only Python >= v3.10 allows single argument calls of `traceback.format_exception`.

In your previous fix, you did not pass the modified traceback `tb`, so `format_exception` internally used the unmodified traceback:
`lines = traceback.format_exception(value)`

bdieterm:
I already tested this new fix with different Python versions from 2.7 to 3.13, and everything worked as expected.

@almarklein: Does this fix also work with the `trio` module and your use case, as you mentioned in #1104?

bdieterm:
I did further testing. As far as I have seen, this also works with `trio`.

PR #1109, opened 2024-12-15
add post-mortem debugging for individual sub-exceptions in ExceptionGroups

bdieterm:
If there is an `ExceptionGroup` in the traceback, post-mortem debugging will only go to the code line where the `ExceptionGroup` was raised. With this PR, individual sub-exceptions (and also nested sub-sub-...-exceptions) can be individually selected for post-mortem-debugging.

The magic command for post-mortem is still `db start`, and this is still the default action when pressing Ctrl+P or selecting "Postmortem" in the "Shell" menu.

This PR changes the following things:
When starting a post-mortem, Pyzo will now detect ExceptionGroups. With no extra argument to `db start`, the first sub-exception that is not an ExceptionGroup will be selected for post-mortem debugging.
And the normal `db start` will inform the user that there are ExceptionGroups and how they can be selected, e.g.:
```
>>> DB START
There is at least one ExceptionGroup. To select the traceback of a specific group,
stop the debugger and type the magic command DB START with (nested) group numbers.
Possible commands for the last traceback are:
db stop         <-- stops the debugger
db traceback    <-- prints the last traceback again
db start        <-- this will automatically choose the first non-ExceptionGroup
db start 0      <-- zero means: explicitly choose the root ExceptionGroup
db start 1      <-- currently used (automatically chosen -- first non-exc-group)
db start 2      
db start 3      
db start 3 1    
db start 3 2    
db start 3 2    

(my_func_b)>>> 
```
The user can then continue debugging in the automatically selected traceback, or stop debugging and pm-debug in another sub-exception, e.g. via magic command `db start 3 1`. The numbers are the ones from the ascii-drawing of the exception-group tree from the traceback. To have a look again at the full traceback with these numbers, the new magic command `db traceback` can be used.

The old behavior of selecting only the root exception group is still available via `db start 0`.
Post-mortem debugging with non-ExceptionGroup errors is not affected by this PR.

Here is some example code that I used for testing when implementing this:
```python3


import sys
import traceback

def my_func_a(x):
    return my_func_b(x)

def my_func_b(y, a=1):
    return a / y

def my_func_c(k):
    raise KeyError('test')

def my_sub_func():
    collected_exceptions = []
    try:
        a = my_func_a(0)
    except Exception as e:
        collected_exceptions.append(e)

    try:
        a = my_func_c(0)
    except Exception as e:
        collected_exceptions.append(e)

    raise ExceptionGroup('sub', collected_exceptions)


def my_func_main():
    collected_exceptions = []
    try:
        a = my_func_a(0)
    except Exception as e:
        collected_exceptions.append(e)

    try:
        a = my_func_c(0)
    except Exception as e:
        collected_exceptions.append(e)

    try:
        my_sub_func()
    except Exception as e:
        collected_exceptions.append(e)

    # raise ExceptionGroup('exception group example', [ValueError(1), TypeError(2)])
    raise ExceptionGroup('main', collected_exceptions)


my_func_main()
```

@almarklein: This also works with the `trio` module, by the way:
```python3
import trio

async def child1():
    print("  child1: started! sleeping now...")
    await trio.sleep(1)
    print("  child1: exiting!")

async def child2():
    print("  child2: started! sleeping now...")
    x = 1 + 2
    raise ValueError("aaa")
    await trio.sleep(1)
    print("  child2: exiting!")

async def parent():
    print("parent: started!")
    async with trio.open_nursery() as nursery:
        print("parent: spawning child1...")
        nursery.start_soon(child1)

        print("parent: spawning child2...")
        nursery.start_soon(child2)

        print("parent: waiting for children to finish...")
        # -- we exit the nursery block here --
    print("parent: all done!")

trio.run(parent)
```

### DIFF
--- a/pyzo/pyzokernel/debug.py
+++ b/pyzo/pyzokernel/debug.py
@@ -339,14 +339,132 @@ class Debugger(bdb.Bdb):
             else:
                 print("Unknown debug command: %s" % name)
 
-    def do_start(self, arg):
-        """Start postmortem debugging from the last uncaught exception."""
+    def _get_traceback_combinations(self):
+        """get all possible combinations for tracebacks
 
-        # Get traceback
+        If the exception is an ExceptionGroup, it can consist of sub-exeptions
+        and even nested ExceptionGroup objects. We identify a branch via the
+        sub-exception numbers that are printed in the traceback ("---- 1 ----" etc.).
+
+        A tuple (3, 2) addresses the ValueError sub-sub-exception in the following example.
+        ExceptionGroup  --> ()
+            ---- 1 ----  --> (1,)
+            ZeroDivisionError
+            ---- 2 ----  --> (2,)
+            KeyError
+            ---- 3 ----  --> (3,)
+            ExceptionGroup
+                ---- 1 ----  --> (3, 1)
+                TypeError
+                ---- 2 ----  --> (3, 2)
+                ValueError
+
+        return value: tb_combs, first_nongroup
+            tb_combs ... a dictionary of kv-pairs
+                key: the tuple branch-identifier
+                value: the traceback of that (sub-)exception
+            first_nongroup ... the key to the tb_combs dict to the first traceback of an
+                non-ExceptionGroup -- useful as a default selection
+        """
+        tb_combs = {}
+        first_nongroup = None
         try:
+            value = sys.last_value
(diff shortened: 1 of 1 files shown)


almarklein:
🤯  Pyzo's debug capabilities are become epic ❤️

PR #1114, opened 2024-12-31
add support for externally started shells

bdieterm:
This PR makes it possible to use externally started Python shells inside Pyzo.

With externally started shells, the Python interpreter of the shell is not started by Pyzo but outside Pyzo by a different application.
In the external shell, some custom start code for the Pyzo kernel has to be executed to create a shell in Pyzo. The Pyzo shell can be closed and re-opened again inside the same Python session. There is no need to add or modify any shell configuration in the "Shell configurations" dialog, since all the relevant information is part of the custom start code.
There are no changes in Pyzo's user interface.

Why could this be useful?
I have a use case where a Python interpreter is started from a third party application that lacks Pyzo's features such as post-mortem debugging and cell code execution.
As a more general example, I added a short description on how to use this with the open source software Blender.

See the info string on the top of file
pyzo/resources/external_shell_examples/start_external_shell.py
for details.

### DIFF
--- a/pyzo/core/commandline.py
+++ b/pyzo/core/commandline.py
@@ -87,6 +87,19 @@ def handle_command(command, arg):
         pyzo.callLater(pyzo.main.close)
         return "Closing Pyzo"
 
+    elif command == "startexternalshell":
+        import ast
+        try:
+            shell_config = ast.literal_eval(arg)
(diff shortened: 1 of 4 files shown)


almarklein:
🤯

PR #1122, opened 2025-01-10
Replace pyproject.toml

almarklein:
Will do ruff-ing in new pr


almarklein:
CI wat?

pyproject.toml
```diff
@@ -0,0 +1,84 @@
...
+# * Update freeze logic if necessary. Merge the pr.
+# * `git tag vx.y.z` and `git push vx.y.z` (builds the binaries and pushes to a GH release)
+# * Update links on Pyzo website
+# * `flit publish`
+
+# ===== Project info
+
+[project]
+dynamic = ["version"]
+name = "Pyzo"
```
bdieterm, line 16: change `"Pyzo"` to `"pyzo"` and then it will work. The name is the module name and is case sensitive.

almarklein: Thanks, that was it!

PR #1124, opened 2025-01-11
restore compatibility with Python 2.7 shells

bdieterm:
Recent commits made Python 2.7 kernels crash in Pyzo.

#1121 removed the `"# -*- coding: utf-8 -*-"` headers from files executed by the kernel.
#1123 used the Python 2.7 keyword `exec` as an attribute, resulting in a SyntaxError

All files inside the folder `pyzo/pyzokernel` are executed only by the kernel (Python 2.7).
All files inside the folder `pyzo/yoton` are executed by both the Pyzo IDE (Python >= v3.6) and the kernel (Python 2.7).
All remaining Python files are executed only by the Pyzo IDE (Python >= v3.6).

I changed all Python scripts that are executed by the kernel so that they are (7 bit) ASCII files.
And I fixed the `exec` attribute and added an override for `ruff` to not touch that anymore.

### DIFF
--- a/pyzo/pyzokernel/guiintegration.py
+++ b/pyzo/pyzokernel/guiintegration.py
@@ -497,7 +497,10 @@ class App_qt(App_base):
                     except Exception:
                         pass  # tough luck
 
-                theApp.exec = theApp.exec_
+                # "exec" as an attribute would cause a syntax error in Python 2.7
+                # because it is a keyword.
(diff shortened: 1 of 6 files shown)


almarklein:
Ai, sorry for forgetting about Python 2.7. Thanks for the superb fixes!

bdieterm:
no problem

Issue #744, opened 2021-03-15
Progress bars are not correctly displayed in shell


MartinGraive:
Hi,
It looks like progress bars in the Pyzo interpreter (using IPython or not) are not behaving like they should be.
Either the bars are not displayed, this is the case when installing (and more precisely downloading) a package with pyzo magic command `pip`; or the bars are displayed one after the other at each increment, as is the case when installing a package with IPython magic command `%pip`. The second case also seems to happen when using package `tqdm`. 
EDIT: For `tqdm`, it sometimes fails completely (one new bar displayed for each progress update); sometimes works well but then prints the completed bar right after the first one, and some other times just does what it should be doing. 
Thank you for this great IDE!

Version info:
Pyzo version: 4.11.2 (source)
Platform: linux
Python version: 3.9.2
Qt version: 5.15.2
PyQt5 version: 5.15.3

almarklein:
Thanks for reporting this issue!

This would need some investigation. Pyzo's shell has *some* terminal emulation, but it is not complete at all. So a progress bar might work. Or it might not :) Depending on how the progress bar works, it might be possible to extend the terminal emulation support to fix it.

bdieterm:
fixed

PR #1129, opened 2025-01-15
improve terminal emulation in Pyzo's shell widget

bdieterm:
Before this PR, the terminal emulation in Pyzo had some flaws when text fragments were not processed in one go.
For example, there were sporadic problems with color format escape sequenes not working or sudden newlines instead of carriage returns. Line shortening for lines > 1024 chars sometimes printed longer lines on the text fragment boundaries.

I completely reworked the terminal emulation.
The best way to get a feeling for the new features is to execute the included example code script cell by cell:
https://github.com/pyzo/pyzo/blob/main/pyzo/resources/code_examples/pyzo_terminal_emulation.py

This also solves issue #744 from the year 2021.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -72,7 +72,7 @@ include = ["README.md", "LICENSE.md", "pyzo.appdata.xml", "pyzolauncher.py"]
 
 [tool.ruff]
 line-length = 88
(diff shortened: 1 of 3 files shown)


almarklein:
wow, nice work!

Issue #752, opened 2021-06-27
Support relative filename as cli argument


jayvdb:
```
pyzo ./src/jinja2/filters.py
Try opening file './src/jinja2/filters.py'
```
Then the GUI says it failed to open `/src/jinja2/filters.py`

```
pyzo --help
Try opening file '--help'
```

almarklein:
I suspect that the filename needs to be absolute (at the moment). Could you try to verify?

bdieterm:
Relative paths are now supported.

PR #1133, opened 2025-01-17
support relative filepaths in Pyzo's command line arguments to open files

bdieterm:
This PR adds support to open scripts in Pyzo by passing a relative filepath as the command line argument.

Before this PR, the command `pyzo ./abc.py` passed the relative path `./abc.py` to the Pyzo command server. The command server then resolved the relative path using the server's current path but not the client's current path.

With this PR, Pyzo resolves the relative path to an absolute path in the client code and then passes the absolute path to the server.
A relative path has to start with a dot, and the `open` command must not be used.
`pyzo ./abc.py` will work as intended now, but `pyzo open ./abc.py` would use the server's working directory.

@almarklein: Can you please test this on macOS because I also changed the `psn_` stuff?

This solves issue #752.

### DIFF
--- a/pyzo/core/commandline.py
+++ b/pyzo/core/commandline.py
@@ -116,11 +116,16 @@ def handle_cmd_args():
     otherwise.
     """
     args = sys.argv[1:]
-    request = " ".join(arg for arg in args if not arg.startswith("--"))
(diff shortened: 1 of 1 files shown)


almarklein:
Works!

PR #1134, opened 2025-01-18
remove remnants from Python <= v2.6

bdieterm:
It has been a while since ancient Python versions such as v2.4 are not supported anymore by Pyzo.
I removed some remnants in Pyzo's source code.

The real minimum supported Python version for Python kernels in Pyzo is still v2.7.

@almarklein: Can you please update the version info on the pyzo.org website?


almarklein:
> Can you please update the version info on the pyzo.org website?

Done

PR #1137, opened 2025-01-20
Fix new tab selector for MacOS

almarklein:
#1135 is awesome but it did not work on MacOS. This fixes that.

### DIFF
--- a/pyzo/core/baseTextCtrl.py
+++ b/pyzo/core/baseTextCtrl.py
@@ -8,12 +8,16 @@ is common for both shells and editors.
 
 import pyzo
 import os
+import sys
 import time
 from pyzo.core.pyzoLogging import print
 import pyzo.codeeditor.parsers.tokens as Tokens
 from pyzo.codeeditor import CodeEditor
 
 from pyzo.qt import QtCore, QtGui, QtWidgets
+
+
+ismacos = sys.platform.startswith("darwin")
 
 
 def normalizePath(path):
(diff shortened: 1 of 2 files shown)


pyzo/core/baseTextCtrl.py
```diff
@@ -656,18 +663,30 @@ def keyPressEvent(self, event):
...
         )
-        if has_control and event.key() == QtCore.Qt.Key.Key_Space:
+
+        # Leave tab events to editor tabs
+        if has_control_like and event.key() in (
+            QtCore.Qt.Key.Key_Tab,
+            QtCore.Qt.Key.Key_Backtab,
+        ):
+            event.ignore()
+            return False
```
almarklein, line 680: Weird that the behavior on Mac and Windows is so different. Also the autorepeat thing ...

bdieterm: This is weird indeed. In Linux and MS Windows this already worked fine before your fix. The widget would then pass the event it to its parent widget.
I guess for macOS, there is some inherited extension (see folder pyzo/codeeditor/extensions) that consumes the keypress when calling `super().keyPressEvent(event)`, because 'event.ignore()' will directly propagate the event to the editor's parent widget which bypasses the event handlers of the remaining extensions. Maybe this `event.ignore()` is just a band-aid for the actual problem somewhere else.

Regarding the auto repeating:
I have no idea why you dismissed auto-repeated events for the pressed Tab-key (while having Ctrl pressed). In Linux and Windows this works fine. A normal press and release fires just a single event. And if you keep Tab pressed longer on purpose, then you get quicker to the desired list entry. I just tried this with a list of 15 entries. Without auto repeating (as now after your fix), pressing and releasing the Tab key more than 10 times in a row is uncomfortable. I use standard keyboard settings, by the way. If this is a macOS thing, we could add a condition to only ignore repeated events for macOS computers.

almarklein: Thanks for testing it out.

Good point about the autorepeat. The behaviour with autorepeat (i.e. without my `if`  to disable it) is the same behavior as e.g. VS Code. I think my ctrl-tabbing flow was a bit confused because it took me a while to get it working again, and for some reason I kept holding tab and it just span around 😄. Will restore.

PR #1139, opened 2025-01-20
fix tabswitching

bdieterm:
@almarklein:
Consider this PR a whiteboard for fixing tabswitching.
I added some commits that show what has to be fixed.

The first commit [use workaround only for macos](https://github.com/pyzo/pyzo/commit/aa14d4f97e7c113b8f5ef4d769482d9181f656f6) is more or less a placeholder for a proper fix.
I think, the problem can be quickly found by deactivating all the `keyPressEvent` methods in the folder "pyzo/codeeditor/extensions" -- just comment out the lines and add a `pass` instead. At least, this is how I did it when I fixed that for Linux.

Commit [fix return statement for PyQt6](https://github.com/pyzo/pyzo/commit/5d978f1b1318c14516b317e3992f3303f6199e7a) is necessary because PyQt6 throws an exception if there is a value returned from the keyPressEvent.

There was also a `print` call from the previous fix. I guess this is not necessary anymore.

The last commit is something I wanted to change anyways: Up to Pyzo v4.19.0, the tab switching key combination was application wide (only the Ctrl+Tab but not the Ctrl+Shift+Tab). With the new tab switching dialog, the key combination only worked when the focus is in the editor(tabs) widget, but not when the focus is in the shell or e.g. filebrowser search textbox. This commit makes the key combinations for tabswitching application-wide again, as before.

### DIFF
--- a/pyzo/core/baseTextCtrl.py
+++ b/pyzo/core/baseTextCtrl.py
@@ -684,7 +684,7 @@ class BaseTextCtrl(CodeEditor):
             )
         ):
             event.ignore()
-            return False
+            return
 
         # Invoke advanced autocomplete/calltips Ctrl+Space key combination?
         if has_control_like and event.key() == QtCore.Qt.Key.Key_Space:
--- a/pyzo/core/baseTextCtrl.py
+++ b/pyzo/core/baseTextCtrl.py
@@ -672,9 +672,16 @@ class BaseTextCtrl(CodeEditor):
         )
 
         # Leave tab events to editor tabs
-        if has_control_like and event.key() in (
-            QtCore.Qt.Key.Key_Tab,
-            QtCore.Qt.Key.Key_Backtab,
+        # FIXME: this is probably just a workaround for another problem
+        # FIXME: ... somewhere in a keyPressEvent in a codeeditor extension
+        if (
+            ismacos
+            and has_control_like
+            and event.key()
+            in (
+                QtCore.Qt.Key.Key_Tab,
+                QtCore.Qt.Key.Key_Backtab,
+            )
         ):
             event.ignore()
             return False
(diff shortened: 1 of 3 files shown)


almarklein:
> I think, the problem can be quickly found by deactivating all the keyPressEvent methods in the folder "pyzo/codeeditor/extensions" -- just comment out the lines and add a pass instead.

I had a stab at this, but the problem persists even when disabling all `keyPressEvent` in the codeeditor subpackage.

I'm starting to think its the text control itself, which is not that surprising since Tab is a key that is normally consumed by an editor. The surprising bit is that it's not on Windows/Linux.

almarklein:
I agree to all the changes in the current PR.

bdieterm:
The following code can be used for testing the key events:
```python3
from PySide6 import QtGui, QtCore, QtWidgets

class MyWidget(QtWidgets.QPlainTextEdit):
    def keyPressEvent(self, event):
        print('keyPressEvent in', self.__class__.__name__, event)
        super().keyPressEvent(event)

class MyGui(QtWidgets.QMainWindow):
    def __init__(self):
        super().__init__()
        layout = QtWidgets.QVBoxLayout()
        self._w = MyWidget('abc')
        self.setCentralWidget(self._w)
        self.show()
        self.activateWindow()

    def keyPressEvent(self, event):
        print('keyPressEvent in', self.__class__.__name__, event)
        super().keyPressEvent(event)

# app = QtWidgets.QApplication([])
gui = MyGui()
# sys.exit(app.exec())
```
On Linux and Windows, the Ctrl+Tab and Ctrl+Shift+Tab are handled by both `keyPressEvent` methods.

@almarklein: By the way, thanks for fixing tab switching for macOS.

almarklein:
> The following code can be used for testing the key events

Works fine; get the event in both the widget and the gui 🤔

PR #1146, opened 2025-01-30
fix filebrowser regexp search and add "whole words" option

bdieterm:
This PR fixes some issues of the filebrowser tool:

When doing a case insensitive search and using a RegExp pattern as the needle, the needle was converted into lowercase, and this turned `\B` into `\b` which has a different meaning in regular expressions. --> fixed

When invoking "Edit -> Find via File Browser" (or Ctrl+Shift+F), the selected text in the editor is searched via the filebrowser tool. But the problem was that this would not always work when option "RegExp" was activated. Now, the needle from the editor will automatically be escaped into a regular expression in the filebrowser search text box.

And this PR adds a new feature:
Now there is a new "Whole words" search option in the filebrowser tool with same behavior as in the code editor's search.


almarklein:
🚀 ❤️

PR #1154, opened 2025-02-09
rework shell configuration dialog

bdieterm:
This is a rework of the shell configuration dialog to improve usability.

The tab container is replaced with a stacked widget and a list. The list selection switches the visible config form (instead of the tab bar). This is more comfortable when having many shell configurations.
Reordering via drag'n'drop is possible (as before with dragging tabs). Duplicating of configurations via drag'n'drop while pressing the Ctrl key is now possible (which was not possible before).

The dialog window has now a larger default size so that everything can be seen without scrolling or resizing.

In the old version, some buttons had weird positions.
"Add config" was on the top right corner in the tab bar widget.
"Delete" and "Copy" were at the bottom of the QForm and not visible without scrolling or resizing of the dialog.
These buttons are now below the new config list widget in the lower left corner of the dialog (outside the form and thus always visible).

Button "Copy" is renamed to "Duplicate".

Window resizing is now done before the dialog is shown so that there is no more flickering of the dialog window while opening.

The "startupScript" text box has no more height limitation.

Large text boxes "pythonPath", "startupScript" and "environ" now have QPlainTextEdit instead of QTextEdit so that there is no more undesired mixed formatting when pasting text.

I will keep this PR a draft for a few days for further testing before merging it.

### DIFF
--- a/pyzo/core/shellInfoDialog.py
+++ b/pyzo/core/shellInfoDialog.py
@@ -37,7 +37,7 @@ class ShellInfo_name(ShellInfoLineEdit):
         self.onValueChanged()
 
     def onValueChanged(self):
-        self.parent().parent().parent().setTabTitle(self.getTheText())
+        self.parent().parent().parent().setShellconfigTitle(self.getTheText())
 
 
 class ShellInfo_exe(QtWidgets.QComboBox):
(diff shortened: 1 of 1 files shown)


almarklein:
Really nice!

PR #1159, opened 2025-03-08
use file encoding information according to PEP 263

bdieterm:
This PR fixes some weird behavior and flaws of Pyzo regarding handling of file encoding and BOM by the editor.

**The current situation (before this PR) is as follows:**
* When opening a file, the encoding is either utf-8 or the encoding specified via the [PEP-263 magic comment](https://peps.python.org/pep-0263/#examples).
* Invalid bytes that are encountered during decoding are silently replaced with a special character (�) -- there is no indication to the user about that, not even in the Logger tool, and the document is marked as unmodified.
* Reloading the file (e.g. after a change of the file modification date) always uses utf-8 encoding even it the file was originally opened with a different encoding. Decoding errors then abort the reloading, and the only indication about that is an exception in the Logger tool and a short message in the status bar (if enabled).
* The encoding selection in the menu (via File -> Encoding) is mostly useless. It displays the encoding from the PEP-263 magic comment, and this reflects the encoding that will be used when saving the file, but that is all. When changing that encoding manually, then saving the file and reopening or reloading that file, that encoding selection is lost. That menu cannot be used to override the encoding that should be used for opening a file.
* When saving a file, a modified or added PEP-263 magic comment will be ignored -- the encoding used will be the one that was determined when opening the file or the one from manually changing via the menu.
* The byte order mark for UTF-8 is allowed according to PEP-263, but Pyzo will remove that BOM when saving changes to the document.

**This PR changes the handling of encodings and byte-order-marks:**
The encoding of a loaded or a newly created file is not a state anymore.
The only thing the editor has to remember about encoding is if there was a BOM_UTF8 in the loaded file.

When loading text from a file, the encoding is determined as follows, similar to PEP 263:
    1) If there is a known BOM, the encoding is defined by the BOM.
    2) Otherwise, if there is a magic comment with a known encoding name,
       then this encoding name will be used.  e.g.: `# -*- coding: latin-1 -*-`
    3) Otherwise utf-8 will be used.

If decoding fails, other encoding candidates from 2) or 3) will be tried as well.
If decoding still fails, ascii with backslashreplace will be used, a warning will
be displayed, and the document will be marked as modified after loading.
The warning consists of a message box and a log entry.

And when saving text to a file:
    1) If there is a magic comment with a known encoding name in the unsaved text,
       that encoding will be used for saving the file.
    2) Otherwise utf-8 encoding will be used.

If there was a BOM_UTF8 present when originally loading the file, that BOM will
only be written to the file if the encoding for saving is utf-8 or utf-8-sig.
Other BOMs, such as BOM_UTF16_LE, will always be discarded when writing a file (because of PEP-263).

I will merge this PR in a few days.

### DIFF
--- a/pyzo/core/editor.py
+++ b/pyzo/core/editor.py
@@ -16,8 +16,70 @@ from pyzo.qt import QtCore, QtGui, QtWidgets
 from pyzo.codeeditor import Manager
 from pyzo.core.menu import EditorContextMenu
 from pyzo.core.baseTextCtrl import BaseTextCtrl, normalizePath
-from pyzo.core.pyzoLogging import print  # noqa
+from pyzo.core.pyzoLogging import print
 import pyzo
+
+
+"""
+Handling of encodings and byte-order-marks
+==========================================
+
+When loading text from a file, the encoding is determined as follows, similar to PEP 263:
+    1) If there is a known BOM, the encoding is defined by the BOM.
+    2) Otherwise, if there is a magic comment with a known encoding name,
+       then this encoding name will be used.  e.g.: "# -*- coding: latin-1 -*-"
+    3) Otherwise utf-8 will be used.
+
+    If decoding fails, other encoding candidates from 2) or 3) will be tried as well.
+    If decoding still fails, ascii with backslashreplace will be used, a warning will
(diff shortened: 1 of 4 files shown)


almarklein:
Nice work!

PR #1185, opened 2025-06-04
SourceSctructure font change and comment cell display

MrXerios:
# Changes
- Change font of source structure tool to match codeeditor
- force update of source structure tool when font of code editor changes
- update cell display text to be of length 120 at least to align the underlining.

# Motivation
This change was motivated by the fact that the underlining isn't justified in the source structure tool for comment cells, which is quite visible in files with a lot of cell comments. This stems from two sources:
1. A fixed pad of 120 whitespaces after the cell name
2. The use of a non-monospace font (sans serif, I think).

Therefore, I propose a change that forces the font of the source structure to match the font of the code editor (which is always monospace), and pad the name of the cell comment to 120 rather than add 120 whitespaces.

### DIFF
--- a/pyzo/tools/pyzoSourceStructure.py
+++ b/pyzo/tools/pyzoSourceStructure.py
@@ -176,6 +176,7 @@ class PyzoSourceStructure(QtWidgets.QWidget):
         prevEditor = self._getCurEditorFromRef()
         if prevEditor is not None:
             prevEditor.cursorPositionChanged.disconnect(self.callbackPosChanged)
+            prevEditor.fontChanged.disconnect(self.updateStructure)
         self._curEditorRef = None
 
         # Get editor and clear list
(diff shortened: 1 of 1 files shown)


bdieterm:
I am fine with these changes.

Just some nitpicks:

1. In method `onEditorsCurrentChanged` please also disconnect the `fontChanged` signal, the same way it is done for `cursorPositionChanged`. Otherwise the callback would be called multiple times for each editor (tab), and even for editor tabs that have been closed again.
2. Maybe you can add another comment that the intention of padding the cell name is to have the underline along the full width.

The format checker "ruff" in the GitHub workflow is quite strict. But it is easy when you use the same checker tool on your computer:
Install `ruff` as a Python package (or via `apt install python3-ruff` on a Debian based Linux). Then cd into the base directory of the Pyzo source code (where file pyproject.toml resides) and run:

`ruff format` ... this will automatically modify the source code files to have the correct format

`ruff format --check` ... this only checks the format without modifying anything

`ruff check` ... this does some code checks using a linter

PR #1187, opened 2025-06-06
Modify ruff commit hook

MrXerios:
Hi,

# Description

Modify the ruff github workflow to:
1. Format automatically rather than check if the formatting is done,
2. Correct linting error that ruff can fix automatically,

This basically runs ``ruff check --fix`` then ``ruff format`` on all files of a PR, then automatically commit any changes.

# Motivation

Mostly my formatting woes from #1185.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,29 +5,46 @@ name: CI
 on:
   workflow_dispatch:
   push:
+    branches: [ main ]
   pull_request:
     branches: [ main ]
 
 jobs:
-  auto-ruff:
+
+  lint-build:
+    name: Fix linting and format errors
     runs-on: ubuntu-latest
 
+    permissions:
+      # add write permissions for auto-commit after ruff changes
+      # added or changed files to the repository.
+      contents: write
+
+    strategy:
+      fail-fast: false
     steps:
     - uses: actions/checkout@v4
-
-    - name: Setup python (v3.13)
+      with:
+        ref: ${{ github.head_ref }}
+    - name: Set up Python
       uses: actions/setup-python@v5
       with:
         python-version: '3.13'
-    
-    - run: pip install ruff
-    - run: ruff check --fix .
-    - run: ruff format .
-    
+    - name: Install dev dependencies
+      run: |
+          python -m pip install --upgrade pip
+          pip install -U ruff
+    - name: Ruff lint
+      run: |
+          ruff check --fix --output-format=github .
+    - name: Ruff format
+      run: |
+          ruff format .
+
     - name: Commit any changes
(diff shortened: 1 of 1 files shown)


bdieterm:
You can and should test this before by creating a new (private) repo on GitHub (with a different name) and uploading a copy of the files in the Pyzo repo there. This is how I test larger changes on the workflows.

I also suggest to incorporate the astral-sh/ruff-action directly into the existing "lint-build" job in the ci.yml file. Otherwise it is not guaranteed that ruff.yml is always run before ci.yml.

MrXerios:
@bdieterm thank you for the tip, It help me not to spam this anymore than I already did. I'm now back with a more polished version of my previous PR. I edited my first message to add more information.

MrXerios:
Hi,

Yes, this has been tested and it does work. I would never ask someone else to do tests for me, though it is appreciated if they do and/or find something I missed.

About the non-related changes,
- the missing name was a mistake ; I just put it back,
- the change to ``push:`` is to allow any branches to be formatted when pushed instead of just the main branch.

In the end, this change only impacts the development side of pyzo. Seeing as you are the main contibutor by a wide margin, feel free to close this PR without merging if you do not like the change. I don't think I'll spend anymore time on this.: my main goal was to learn the basics of github actions, and that has been accomplished.

bdieterm:
@MrXerios:
I tested your version on a temporary repository that I created only for your PR.
The git-auto-commit action did not work, though, because it had no write access to add a commit.
I have no idea how this worked for you without other modifications of the workflow.
The permission stuff is described at the [main page of the action](https://github.com/stefanzweifel/git-auto-commit-action#usage), by the way.
There was also the `head_ref` setting missing that is necessary for `pull_request` events, according to the [description](https://github.com/stefanzweifel/git-auto-commit-action#example-workflow).
I added a commit to your PR branch to fix these problems. Unfortunately, the CI workflow fails because `/usr/bin/git ... fetch ...` runs into a timeout, even though that ci.yml worked fine on my test repo.


> I don't think I'll spend anymore time on this.

I also think that I will not spend more of my spare time on this, given that this is only a conveniene feature for rare cases. I am not very familiar with GitHub workflows anyways.
First time contributors would not see the auto-commits in PRs until the workflow run is manually approved by a Pyzo maintainer. And when doing the manual approving, we could add a comment on how to fix the formatting like I did with your first PR. Next time I will do this immediately when approving the workflow. We could also add a sentence about this in the contributions section in the README.md file. And we can also manually add a commit to PRs to fix the ruff issues.
For me, the added complexity is not worth it to invest much more time into this topic.
But I am not opposed to that feature as long as it is implemented properly without causing other problems.
If you want, you can improve your knowledge by digging deeper into this topic. Or maybe @almarklein wants to spend time on this feature.

Issue #1172, opened 2025-04-17
Enhancement: synthax highlighting for formatted strings


MrXerios:
Hi,

I noticed that the text highlighting for formatted strings does not take into account the pieces of code embedded, which makes them way harder to read in my opinion.

Example:
<img width="498" alt="Image" src="https://github.com/user-attachments/assets/dbcf115f-d714-47d7-8078-4ac6aef0c8df" />

my proposal: change how the highlighting works so that embeded code (here 1 + 1) is highlighted like it is out of a string (here <img width="40" alt="Image" src="https://github.com/user-attachments/assets/1bd11578-6d37-449b-bff7-a833c3e3574c" />) and maybe add a specific highlingting for formatting directions (here :.2f ).

Have a nice day,
Greg

MrXerios:
update: I have been looking into a solution, will submit a PR when I have something working. Out of curiosity, why use a custom parser instead of a well maintained parser such as pygments ?

bdieterm:
> Out of curiosity, why use a custom parser instead of a well maintained parser such as pygments?

One of the main features I like about Pyzo is that is has no dependencies other than a single Qt wrapper (PySide2/6 or PyQt5/6).
Compare that to [more than 48 packages required by SpyderIDE](https://github.com/spyder-ide/spyder/blob/master/requirements/main.yml).

bdieterm:
I am aware of the shortcomings of the current parser in Pyzo.
For my own use, this is no problem at all because when using f-strings, I actually only have simple cases without nesting, for example:
```python3
x = 123; print(f'{x = }')
y = 45.67; print(f'{y = :.3f}')
z = -89; print(f'z: {z}')
```
My spare time is limited, so I am not highly motivated to rework the current parser. I once had a quick look into using `tokens = ast.parse(code_text)` for that, but that only works for full code that could also be compiled without syntax error.
If you want to improve syntax highlighting in Pyzo and maintain that code and related code for the years to come, feel free to create a PR.

Issue #762, opened 2021-10-29
Method names not recognized when using non-ASCII characters


alitekdemir:
Function names do not appear in bold when I use Turkish characters (çöşığüÇÖŞİĞÜ).
it is also not listed in the Source structure field.
Sample code
```python
def kisiler(self):
    print('Personel listesi:')
    for kişi in self.personel:
        print(kişi)
def kişiler(self):
    print('Personel listesi:')
    for kişi in self.personel:
        print(kişi)
```
![pyzo_nLybU3n8qn](https://user-images.githubusercontent.com/1507344/139473678-93bf6923-3277-4262-9dfc-f0e472ac0bc3.png)

almarklein:
Thanks for reporting this! It looks like the regexes only cover ASCII chars.

PR #1195, opened 2025-07-03
Python parser rework

MrXerios:
Hi,

This is a pretty big PR focused on making a better python syntax parser.

**While this has been tested thouroughly, there are probably still some issues to find. If this is merged, I'll check the github issues often for the next while to correct any arising bug as fast as possible.**

# Motivation

The main motivation of this PR is to make syntax highlighting for f-strings (and now t-strings) a bit easier to read. Since this was actually quite tricky to add to the previous parser, I decided to simply implement a new one.

I also took advantage of this change to implement some other changes, which a described in more details later in this PR.

# Description

## 1°) Overview
This PR implements a new subclass of parser, the `RegexParser`, which is heavily inspired by Pygments. (see [https://pygments.org/docs/lexerdevelopment/#regexlexer](url)). 

A quick overview of this type of pasers is given in a new example file (`pyzo > ressources > code_examples > regex_parser_example.py`)

This new subclass is then used to implement a new `Python3Parser` (and `PythonParser`, and `Python2Parser`)

## 2°) Concrete examples

This is not a 1 for 1 replacement for the previous parser, here are some of the main differences:

### a.) f-string syntax highlighting (see issue #1172)
before:
image
after:
image

### b.) Proper unicode identifiers (see issue #762)
before:
image
after:
image

### c.) todo continuity (allow todo comments to be multiline without writing `todo` again)
before:
image
after:
image

# Known issues

For now, there is only one known edge case where this parser has an incorrect behaviour. This is when `}` is inserted into an f-string expression. 
example:
image

This can be corrected, but I'm not quite sure how to do it elegantly yet.

# Ideas future development

I have a few idea of possible evolutions for pyzo steming from this new parser, or using this new parser subclass. Here they are, in no particular order:
- Add type-hint syntax highlighting,
- Add syntax highlighting for attributes / methods,
- maybe add a decorator highlighting too,
- Token overhaul to make syntax highlighting more customizable,
- Better SourceStructure tool (seriously it is quite broken, it should just use the current parser),
- More syntax highlighting themes. There are great ones out there.

Feel free to give me your opinion of them if you feel like it, or add to the list. I don't guarantee that I'll do all that, but I'll certainly do some of them if the PR is accepted.

(_Now I'll create this pull request and fail every single CI test, I'm calling it._)
_EDIT: yes I did_


bdieterm:
Hi @MrXerios,

contrary to you I only did a bit of testing instead of doing that "thouroughly", and I already found many problems that are a no-go to merge this PR. Please be more meticulous when doing further tests. I would be surprised if you wouldn't encounter some of the problems I already encountered when you use Pyzo normally for some hours or days.
Be aware that the current Pyzo release v4.19.0 has more than 30k downloads since its release half a year ago. So there are many users that would experience problems.


## Examples for bugs

Here are some examples for bugs that I already encountered, and there are probably many more:

#### Bad performance when starting Pyzo with 33 open files (mostly from the Pyzo repo)
Time is measured when starting Pyzo after the splash screen disappears till Pyzo is fully loaded.
With your branch, this took 48 seconds.
With the current main branch, this took only 4 seconds --> your version is more than 10 times slower.
When going through your code changes, I already found some bottle necks. After fixing them I got your branch down to 18 seconds, but this is still much slower. My computer is quite old, by the way.

#### Bad performance while typing in the Pyzo editor
Especially with larger files, Pyzo will lag during typing.
For example when editing baseTextCtrl.py (from the Pyzo repo) and removing the three quote signs in line 7, Pyzo is stuck for a moment. You will notify this even more when you get back the three quotes via Ctrl+Z.
Pyzo main branch is snappy, and this is a feature that Pyzo users like.

#### Match soft keyword does not work anymore
In https://github.com/pyzo/pyzo/issues/1172#issuecomment-2835926595 you mentioned that there are some edge cases where the match statement is not parsed correctly.
But in your branch, even a very simple, common case does not work:
When you enter the line `x = match` where "match" is a variable, then it must be rendered as an identifier and not as a keyword. Even you use the variable name "match" in your PR (see method "find_match").

#### TODO comments do not work anymore
Obligatory whitespace characters after "TODO" and "FIXME" are not checked anymore.
In your branch, a comment line like `# todos los problemas` is rendered as a TODO comment.

#### Matlab/Spyder-like comment cells do not work anymore
`## abc` --> ok
`#%% abc` --> nok
`# %% abc` --> nok

#### Auto indent after a block head with a comment does not work anymore
e.g.: when typing the line `for x in range(10):  # abc` and pressing RETURN, there is no auto indent anymore

#### Auto completion after a string literal (with a dot at the end) does not work anymore

#### Parser gets completely stuck on a very simple input
create new file
press RETURN key (so that you have two lines)
move cursor back to line 1
type `f"{`
--> parser is stuck
This is severe because that means that Pyzo users will lose unsaved data.
When executing Pyzo inside Pyzo, you can get out of this via Ctrl+I in the parent Pyzo shell.

#### F-strings with single quotes do not work (contrary to double quotes)
`f"{1 + 1}"` --> ok
`f'{1 + 1}'` --> nok

#### "Execute line and print result (F10)" is broken

#### Python2 parsing is broken
e.g. `print abc` --> print statement is not bold anymore

#### String literals with escaped characters do not work anymore
`"abc\"def"` --> `def` is part of the string but parsed as a keyword
`"abc\\'def"` --> `def` is part of the string but parsed as a unterminated single quote string literal

#### ... and probably much more


## Code review
These are some notes I made when going through your code, but only shallowly.
The order is the one I got from my diff tool, so a mix of severe bugs and tiny improvements.
There are also the code lines I used to improve the performance.

#### pyzo/codeeditor/extensions/behaviour.py
line 332:
Comment "# tokens[-2] if len(tokens) >= 2 else None" makes no sense anymore and can be removed

line 527:
The check `isinstance(token, MultilineFStringToken)` is actually not necessary because `MultilineFStringToken` is derived from `MultilineStringToken` and therefore also an instance of `MultilineStringToken`.

lines 547-549:
parens instead of brackets (tuple instead of list) would be a little bit better

lines 563 and 565:
both lines can be removed because of isinstance with: `class FStringToken(StringToken):` and `class MultilineFStringToken(MultilineStringToken):`

#### pyzo/codeeditor/highlighter.py:
lines 67-71:
These lines make no sense. I guess you used them to set a breakpoint for debugging, but they should be removed again.

#### pyzo/codeeditor/manager.py:
lines 112-113:
These lines make no sense. I guess you used them to set a breakpoint for debugging, but they should be removed again.

#### pyzo/codeditor/parsers/__init__.py:
line 168:
`class from_list` --> please use CamelCase for class names

line 182:
please use CamelCase for class names

lines 187-189:
This function is used nowhere.

line 194:
instead of `tuple()` write just `()`

line 201:
"A tutorial" --> looks like an unfinished sentence in the doc string.

line 206:
Why did you use the regex `r"$.*\Z"`, out of curiosity?

lines 216-224:
Explanation is more confusing than the code itself.
Also some binary numbers are written with a big letter `O` instead of a zero decimal character `0`.
The meaning of value -1 is not described, and the order (little/big endian) also not.
Reading the code is much clearer in this case.

lines 226 and 229:
the `stack` variable could be always initialized with `["root"]` and returned with a single common `return` instead of multiple return statements.

lines 232-234:
using `divmod` would be more elegant: `block_state_val, n = divmod(block_state_val, N)`

line 297:
"the the" --> "to the"

line 306-307:
The docstring of the function is wrong.
There is also a third value `next_state` that is returned (and not only two values).
If there is no match, then tuple `(None, None, None)` is returned and not just `None`.
typo: "occures" --> "occurs"

line 310:
`pattern` is (or must be) an already pre-compiled regular expression.
--> change `match = re.match(pattern, line[pos:])` to `match = pattern.match(line[pos:])`
This is one of the two parts to improve the performance.

line 317:
Please add the information in the doc string that every case consists of `pattern, token_type, next_state` where the second or second and third value are optional because of the padding with None values.

line 324:
typo: "Inludes" --> "Includes"

line 330:
Patterns are added without compiling them:
change line
`temp[state].extend(case.to_pattern())`
to lines
`for pat, *others in case.to_pattern():`
`    temp[state].append((re.compile(pat), *others))`
This is the second of the two parts to improve the performance.

line 358:
Function `process_includes` is only a part of method `process_states`.
You could move that inside method `process_states` or inside class `RegexParser` as a protected static method.

#### pyzo/codeeditor/parsers/python_parser.py
line 429:
`//=` is missing in the operators list

line 462:
`stringLiteralPrefixes` should remain a `frozenset` for faster membership tests for auto-closing quotes

lines 468-473:
Where are these values taken from?
How do maintainers check if the values are still up to date in some years?
Could there be a license violation?
In case of doubt we will continue to only support ascii identifiers, or a limited set of unicode characters, or too many characters. This does not have to be 100 % accurate.
e.g.
`uni_name = r"[^\d\W]\w*"`
or
`uni_name = r"[a-zA-Z0-9_][a-zA-Z0-9_]*"`

line 488 and 729:
missing alternative cell comments: "#%%", "# %%"

line 539 and 778:
typo: Elipsis --> Ellipsis

line 556:
You don't need the `+` to connect these string literals.

line 637:
typo: lien --> line

lines 648-655:
There is your mentioned bug with the TODO.

#### pyzo/resources/code_examples/regex_parser_example.py
line 5:
typo: RegexParseer --> RegexParser
typo: Pyments --> Pygments

line 62:
change `print(tok.__repr__())` to `print(repr(tok))`

The folder "code_examples" is actually for end users and plug-in developers.
Therefore, the description of the parser should be moved to file pyzo/codeditor/parsers/\_\_init\_\_.py next to class `RegexParser`. You could move the example code inside an `if False:` block followed by a cell comment.

## License and attribution
I am not experienced with copyright stuff.
Is there any attribution missing to not violate any laws and licenses?
For example, who wrote the string literal for `xid_continue` which has more than 11000 characters?

MrXerios:
Hi,

Thanks for that comprehensive review. This is some very valuable feedback. Most of these bugs should be easy enough to solve with some better regex. I'll take a look at those that can't be sometimes this week.

The main issue is the performance hit, which is kind of unavoidable with this structure of parser. I admit I did not try to open that many files at once so the performance issues looked acceptable. You seem to have found some good ways to improve performances already, I'll take a look and try out some of my own ideas. We'll see how it goes.

Issue #289, opened 2014-04-08
Minimalistic view (full-size editor or shell)


almarklein:
Originally reported by: **Anonymous**

---

Matlab IDE has a feature that I uses quite often:
Whenever you double-click the title-bar of a window, it maximizes, and becomes the only window remaining. It is really helpful when you need to place matlab side-by-side with some other texts. 

Please consider adding this feature to IEP, no need to add it to every tools, maximizing the editor and shell would be enough.

I tried to edit the source code myself, and for the editor it seems to works fine, but not for the shell window. Here is what I got so far:

in class EditorTabs:

is_maximized = False

```
def toggleMaximize(self):
    """ Toggle maximize status of the editor window. """
    from iep.tools import ToolManager
    for toolname in iep.toolManager.getLoadedTools():
        tool = iep.toolManager.getTool(toolname) 
        if self.is_maximized:
            tool.parent().show()
        else:
            tool.parent().hide()

    if self.is_maximized:
        iep.shells.parent().show()
    else:
        iep.shells.parent().hide()

    self.is_maximized = not self.is_maximized                    
    self.show()
```

Thank you for the excellent work.

---
- Bitbucket: https://bitbucket.org/iep-project/iep/issue/289

PR #1217, opened 2025-10-11
add feature to manage and switch views

bdieterm:
This PR makes it possible to define multiple views and easily switch between them. A "view" contains the window and panel geometries of the Pyzo IDE, including the tools.

There is a new default keyboard shortcut Ctrl+Shift+A for switching views, similar to the existing Ctrl+Tab for switching between editors.

See the docstring in pyzo/core/views.py for more details.


almarklein:
Ah, zen mode (#289)

PR #1247, opened 2026-02-15
clean-up code in Qt event loop

bdieterm:
@almarklein:
I am wondering why there is a call to the original `QApplication.exec` with an argument.
As far as I know, `QApplication.exec` (and `QApplication.exec_`) take no arguments, so the first `try` will always raise an exception.
I will merge this in a few days, unless there is a reason to keep the current code.


almarklein:
That is strange indeed. I would guess that at some point it got the method of the class (the unbound method), and passing the application object then becomes the same as calling the method on that instance. But the code before clearly takes the method of the instance ...

almarklein (approved):

PR #1251, opened 2026-04-13
Fix datadir for termux

### LINKED ISSUE #1250: unsupported operating system android
i use pyzo on termux on android arm64. with last update of termux i had to get latest version of pyzo (4.21.0), but now pyzo doesn't start anymore with message "unsupported operating system android".

almarklein:
Closes #1250

Todo:
* [ ] Check more instances of `sys.platform`, maybe add a `pyzo.platform` to reuse throughout? Or use `os.name` in cases where we need to distinguish between posix and nt systems.

### DIFF
--- a/pyzo/util/paths.py
+++ b/pyzo/util/paths.py
@@ -17,6 +17,12 @@ def prepare_appdata_appconfig_dirs():
     folders are created if not present
     """
 
+    # Select platform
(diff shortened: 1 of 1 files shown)


bdieterm:
@almarklein
We already merged another fix for the problem. How about closing the PR? Or are you still working on that?

PR #1256, opened 2026-04-26
set default shortcut for stopping debugging to Ctrl+Shift+P

bdieterm:
I have already used this new shortcut for some time and I perceive it as quite comfortable.

When debugging, almost everything is done via the keyboard, like running the script and then entering post-mortem debugging via Ctrl+P.
Before running the script again, the debug mode has to be stopped, and this is where it was always necessary to move the hand to the mouse and to click that icon.
I chose Ctrl+Shift+P as the shortcut, because it is the "shifted" shortcut version of entering post-mortem-debugging.

I will wait a bit with merging this PR in case there are contrary opinions, like possible issues with macOS.


almarklein (approved):
Nice!

PR #1258, opened 2026-05-02
add post-mortem debugging for chained exceptions

bdieterm:
I reworked some of the debugging methods and added a feature for post-mortem debugging:

Exceptions that caused other exceptions can now also be debugged post-mortem.

Here is an example:
```python3
def myfunc1(a, b):
    return a / b

def myfunc2():
    try:
        myfunc1(4, 0)
    except Exception as e:
        raise TypeError('abc')
        # raise TypeError('abc').with_traceback(e.__traceback__)  # common before PEP 3134
        # raise TypeError('abc') from e  # introduced via PEP 3134

        # see:
        #   https://docs.python.org/3/reference/simple_stmts.html#the-raise-statement
        #   https://docs.python.org/3/library/exceptions.html#BaseException.with_traceback

myfunc2()
```

When executing the code, `myfunc1` causes an exception. That exception is caught by the caller `myfunc2`, which then raises another exception which terminates execution.

Starting post-mortem debugging afterwards does not provide us the stack frame from the original zero-division-exception in `myfunc1`.

When using the `raise` variant with the `with_traceback`, we have have another stack frame in post-mortem and can see inside `myfunc1`. But the use of `with_traceback` is not common anymore for indicating the root-cause-exception.

Since PEP 3134, there is a `raise` statement with the `from` keyword. In such a traceback we then see the original exception, then the message "The above exception was the _direct cause_ of the following exception:", and then the exception of the caller.
When we had this "raise new_exc from orig_exc" before post-mortem debugging, Pyzo did not add the stack frames of the original exception (root cause) to the frame list for post-mortem stack frames.
With this PR, these stack frames are now added as well, so we have "deeper" post-mortem debugging when using the raise-from-clause.

### DIFF
--- a/pyzo/pyzokernel/debug.py
+++ b/pyzo/pyzokernel/debug.py
@@ -55,28 +55,24 @@ class Debugger(bdb.Bdb):
 
         return bdb.Bdb.trace_dispatch(self, frame, event, arg)
 
-    def interaction(self, frame, traceback=None, pm=False, framesBefore=None):
+    def interaction(self, frame, traceback_=None, pm=False, pm_exc=None):
         """Enter an interaction-loop for debugging. No GUI events are
         processed here. We leave this event loop at some point, after
         which the control flow will proceed.
 
         This is called to enter debug-mode at a breakpoint, or to enter
         post-mortem debugging.
-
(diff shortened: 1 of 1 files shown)


almarklein:
Supercool, thanks! ❤️

PR #1268, opened 2026-06-05
prevent deletion of prompt and output lines in shell

### LINKED ISSUE #1267: Unexpected interaction between ctrl+backspace and the terminal
I found an interaction that I'm guessing is unexpected between the ctrl+del shortcut, to remove a whole word, and the pyzo terminal. Basically, it's a way to remove the begining of a line in the terminal.

Steps to reproduce:
- Open pyzo and go to the terminal,
- Add any amount of whitespace to the current line, 
- Press ctrl+del,

images of the result:

Usual terminal
Image

Whitespaces are a mix of spaces and tabs, highlighted to make them visible
Image

`>>>` is missing
Image

This is also an issue when debugging:

before:
Image

after 1:
Image

after 2:
Image

etc.

I could reproduce this issue on windows 11 with pyzo 4.21 and on linux with pyzo 4.20.

EDIT: There is also an issue with highlighting instead of whistespace, which probably has the same root cause. This is even more unexpected since it enables deleting previous lines of the terminal:

Image

highlight text:
Image

ctrl+del
Image

reepat 5 times:
Image

bdieterm:
There were three bugs in the shell that resulted in unintended deletion of text in the shell:

1. With the cursor at a fresh prompt, select some characters towards the left and then press Backspace.
--> The space at the end of the prompt is deleted (`>>>` instead of `>>> `).
2. Select some text lines from above towards the end of the prompt (after `>>> `) and then press Backspace.
--> The whole selection, including the prompt is deleted.
3. Enter only whitespace character after the prompt (e.g. `>>>      `) and then press Ctrl+Backspace.
--> The whole line (including the prompt) is deleted.

This fixes #1267.

### DIFF
--- a/pyzo/core/shell.py
+++ b/pyzo/core/shell.py
@@ -29,6 +29,7 @@ from pyzo.core.pyzoLogging import print
 from pyzo.core.kernelbroker import KernelInfo, Kernelmanager
 from pyzo.core.menu import ShellContextMenu
 
+ismacos = sys.platform.startswith("darwin")
 
 # Interval for polling messages. Timer for each kernel.
 POLL_TIMER_INTERVAL_IDLE = 100  # in ms; energy saving mode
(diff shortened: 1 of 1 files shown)


bdieterm:
@almarklein: Could you please check this on a macOS computer.

almarklein:
Looking good 👌

